PR #208, opened 2020-02-04
Rebase 80.0.3987.87

romandev:
This patch includes the following things:
  - Update BASE file to 80.0.3987.87
  - Update upstream base files based on 80.0.3987.87
  - Update diff files based on 80.0.3987.87
  - Update triple_banana changes for 80.0.3987.87

Refs: none


upstream/triple_banana_components_password_manager_core_browser_password_save_manager_impl.cc.patch
```diff
@@ -0,0 +1,40 @@
...
++
+ using autofill::FormData;
+ using autofill::FormFieldData;
+ using autofill::FormStructure;
+@@ -173,6 +179,22 @@ void PasswordSaveManagerImpl::CreatePendingCredentials(
+       pending_credentials_.signon_realm = parsed_submitted_form.signon_realm;
+       pending_credentials_.action = parsed_submitted_form.action;
+     }
++#if defined(ENABLE_TRIPLE_BANANA)
++    // FIXME(zino): We should refactor the following code matching with m80.
```
romandev: @jdragonbae Can you check this logic? The `SetPasswordOverriden()` is no longer available.

src/org/triple/banana/toolbar/BottomToolbarController.java
```diff
@@ -107,7 +107,6 @@ private void makeToolbarMenuButton() {
         View view = View.inflate(BananaContextUtils.get().getApplicationContext(),
                 R.layout.toolbar_menu_button, null);
         mMenuButton = view.findViewById(R.id.menu_button_wrapper);
-        mMenuButton.setWrapperView(view.findViewById(R.id.labeled_menu_button_wrapper));
```
romandev: @codeimpl The setWrapperView was gone in upstream side[1]. Can you check whether we can also remove this?
[1] https://chromium-review.googlesource.com/c/chromium/src/+/1954866

codeimpl: Yes, we can remove it.
Now we don't use menu button because  top toolbar has menu button fixed.

upstream/triple_banana_components_password_manager_core_browser_password_save_manager_impl.cc.patch
jdragonbae: I think we can set the `pending_credentials_state_` instead of `SetPasswordOverriden()`.
Please, refer to #210.

PR #209, opened 2020-02-05
Rebase 80.0.3987.87 bottom toolbar flag

jdragonbae:
This patch enables flag for bottom toolbar.
It enables to use bottom toolbar when first start the app.
In chromium 80, it should be initialized on native side.
    
Refs: #208, #197


jdragonbae:
PTAL

jdragonbae:
Refer: #216

PR #210, opened 2020-02-05
Fix a bug that update password prompt always appears

jdragonbae:
This is a regression since m80 rebase.

Refs: none


romandev:
@jdragonbae lgtm but could you please rebase this quickly?

PR #216, opened 2020-02-05
Fix a bug that bottom toolbar wasn't initialized

jdragonbae:
This is a regression since m80 rebase.

Refs: none


upstream/diff/chrome/browser/android/chrome_feature_list.cc
```diff
@@ -0,0 +1,671 @@
...
+const base::Feature kCCTResourcePrefetch{"CCTResourcePrefetch",
+                                         base::FEATURE_ENABLED_BY_DEFAULT};
+
+const base::Feature kCCTTargetTranslateLanguage{
+    "CCTTargetTranslateLanguage", base::FEATURE_ENABLED_BY_DEFAULT};
+
+const base::Feature kChromeDuetFeature {
+  "ChromeDuet",
+#if defined(ENABLE_TRIPLE_BANANA)
+      base::FEATURE_ENABLED_BY_DEFAULT
```
jdragonbae: @romandev , @codeimpl : PTAL

romandev:
Hmm, We already have a public API (initCommandLine) because we should call initCommandLine() at somewhere.

jdragonbae:
@romandev Thanks for your review. 
@codeimpl , @romandev : PTAL

PR #225, opened 2020-02-10
Fix inform text in ToolbarEditActivity

codeimpl:
Replace bottom toolbar inform text as "drag and drop buttons"

Refs: #182


res/layout/toolbar_edit_main.xml
```diff
@@ -8,7 +8,7 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:clipChildren="false"
-    android:layout_height="match_parent">
+    android:layout_height="match_parent" >
```
romandev: Why do we need a white space here?

codeimpl: done!

Issue #180, opened 2020-01-26
Should update the appropriate state of toolbar button


romandev:
For example, if user clicks `Add Bookmark` toolbar button, should update the state(e.g. `Added`) of the button after adding a bookmark.

PR #258, opened 2020-02-14
Update the state of bookmark toolbar button

jdragonbae:
This patch updates the state of bookmark toolbar button.
    
Refs: #180


banana_cake/interfaces/java/org/banana/cake/interfaces/BananaBottomToolbarController.java
```diff
@@ -30,4 +30,5 @@ void initializeWithNative(OnClickListener tabSwitcherListener,
     boolean isShowingAppMenuUpdateBadge();
     MenuButton getMenuButton();
     void destroy();
+    void updateBookmarkButtonStatus();
```
romandev: If it takes `isBookmarked`, `editingAllowed`, there is no reason to add `isBookmarked()` and `isBookmarkEditingAllowed()`

jdragonbae: Yes. you're right. 
If then, bookmark button update should be implemented in other function like `updateBookmarkButtonState(isBookmarked, editingAllowed)` instead of `updateButtonState()`.
Is it the right way?

src/org/triple/banana/toolbar/ToolbarButton.java
```diff
@@ -131,6 +135,41 @@ public void onUpdateUrl(Tab tab, String url) {
...
+                                : AppCompatResources.getColorStateList(
+                                        BananaApplicationUtils.get().getApplicationContext(),
+                                        R.color.blue_mode_tint));
+            } else {
+                mImageButton.setImageResource(R.drawable.btn_star);
+                ApiCompatibilityUtils.setImageTintList(
+                        getImageButton(), mThemeColorProvider.getTint());
+            }
+            setEnabled(isBookmarkeditingAllowed);
+        }
```
jdragonbae: This is what I questioned above.

banana_cake/interfaces/java/org/banana/cake/interfaces/BananaBottomToolbarController.java
romandev: Yeah, it's better

romandev:
@jdragonbae Please rebase

jdragonbae:
> @jdragonbae Please rebase

@romandev Done! PTAL

src/org/triple/banana/toolbar/BottomToolbarController.java
```diff
@@ -169,6 +169,17 @@ public void onIncognitoStateChanged(boolean isIncognito) {
         mIsIncognitoMode = isIncognito;
     }
 
+    @Override
+    public void updateBookmarkButtonStatus(boolean isBookmarked, boolean editingAllowed) {
+        for (int i = 0; i < MAX_BUTTON_SIZE; i++) {
```
romandev: This is not good. To update the button state, is it correct to make this algorithm take O(n) times for all toolbar buttons? The updateBookmarkButtonState will happen pretty often.

jdragonbae: Thanks for your review. I agree with you.

jdragonbae:
@romandev I'm not sure this patch would be the right way. PTAL!

romandev:
I think this patch should be refactored but let's merge this for now because we don't have enough time to release.

Issue #253, opened 2020-02-13
Need to enable/disable "Enable Bottom Toolbar" feature according to the flag


romandev:

romandev:
@jdragonbae any update? I looked at your patch (non pull request) but its direction is not good. I think we should enable/disable it in runtime (without restarting browser)

jdragonbae:
@romandev Thanks for the guidance. I'll consider it to be possible at runtime.

romandev:
@jdragonbae If it is difficult to change it in runtime, let's do it with restart. I already made utility method in `ExtensionFeatures`. If you don't have a time, I'll take this issue.

PR #282, opened 2020-02-19
Implement BOTTOM_TOOLBAR flag in ExtensionFeatures

jdragonbae:
This flag enables bottom toolbar when triple banana launched.

Refs: #253


res/values-ko/strings.xml
```diff
@@ -15,8 +15,9 @@
     <string name="background_play">백그라운드 재생</string>
     <string name="background_play_summary">탭이 활성화 상태가 아니거나 장치의 화면이 꺼져있을 때도 미디어를 재생하도록 합니다</string>
     <string name="banana_extension">Banana 확장 기능</string>
+    <string name="bottom_toolbar">하단 도구모음 활성화</string>
+    <string name="bottom_toolbar_summary">사용자가 자주 사용하는 기능을 하단 도구 모음에 배치하고 실행 가능하게 합니다\n(예: 뒤로 가기, 탭 추가 등)</string>
     <string name="drag_and_drop_buttons">자주 사용하는 기능을 도구모음에 배치하세요</string>
-    <string name="enable_bottom_toolbar">하단 도구모음 활성화</string>
```
romandev: We don't want the name change because the string means "enable_bottom_toolbar".

res/values-ko/strings.xml
```diff
@@ -15,8 +15,9 @@
     <string name="background_play">백그라운드 재생</string>
     <string name="background_play_summary">탭이 활성화 상태가 아니거나 장치의 화면이 꺼져있을 때도 미디어를 재생하도록 합니다</string>
     <string name="banana_extension">Banana 확장 기능</string>
+    <string name="bottom_toolbar">하단 도구모음 활성화</string>
+    <string name="bottom_toolbar_summary">사용자가 자주 사용하는 기능을 하단 도구 모음에 배치하고 실행 가능하게 합니다\n(예: 뒤로 가기, 탭 추가 등)</string>
```
romandev: Do you think that the more information is necessary?
I thouht that users could understand without further explanation. (Unlike Safe Login, Background Play and so on)

src/org/triple/banana/settings/ExtensionFeatures.java
```diff
@@ -54,6 +54,14 @@ public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
             showRestartDialog();
             return true;
         });
+
+        SwitchPreferenceCompat bottomToolbar =
+                (SwitchPreferenceCompat) findPreference(FeatureName.BOTTOM_TOOLBAR);
+        bottomToolbar.setOnPreferenceChangeListener((preference, newValue) -> {
+            setEnabled("bottom_toolbar_enabled", (boolean) newValue);
```
romandev: Do we really need this?

res/values-ko/strings.xml
jdragonbae: Thanks. I'll change.

res/values-ko/strings.xml
jdragonbae: Thanks! got it. The user seems to think like you said. I'll change.

src/org/triple/banana/settings/ExtensionFeatures.java
jdragonbae: `bottom_toolbar_enabled`. This preference value is referenced by bottom controller[1] to enable bottom toolbar. 

If this code not, it seems like doesn't apply immediately after restart.
Because, reading preference(isBottomToolbarEnabled) is first and writing preference (capture the flag[2]) later. 

This means that `enable-feature` initial command would be not work as we expected.

[1] https://bit.ly/2V7o5ZQ
[2] https://bit.ly/2uRJMCJ

romandev: Set the `FeatureName.BOTTOM_TOOLBAR` to `"bottom_toolbar_enabled"`; Then, we can simply sync with the internal preference value with UI. (We should leave a comment why we have the different style name.)

res/values-ko/strings.xml
romandev: You didn't update this yet.

res/values/strings.xml
```diff
@@ -15,8 +15,9 @@
     <string name="background_play">Background Play</string>
     <string name="background_play_summary">Enable play media in background when tab is not active or device screen is turned off</string>
     <string name="banana_extension">Banana Extension</string>
+    <string name="bottom_toolbar">Enable Bottom Toolbar</string>
```
romandev: ditto

src/org/triple/banana/settings/ExtensionFeatures.java
jdragonbae: Thanks for your review! Done.

res/values/strings.xml
jdragonbae: Sorry for late. Done.

Issue #334, opened 2020-03-15
Bottom toolbar is not shown


romandev:
We are not sure how we can reproduce this issue exactly. But it anyway happened on one device.

PR #341, opened 2020-03-16
Add the logs for bottom toolbar

jdragonbae:
This patch means adding the logs for bottom toolbar. Recently, there is
some issue that bottom toolbar is not shown.
It's difficult to reproduce. By putting the logs at a point related to
bottom toolbar, it can be helpful to find out the problem.

Refs: #334


romandev:
Closed this because it's difficult to apply this for now due to privacy policy.

Issue #377, opened 2020-04-20
A crash when running browser on N OS device


jdragonbae:
This crash seems to happen for the below reason.

- `String.join` cause the crash. Since it is supported only in java8 or higher.

PR #379, opened 2020-04-20
Use TextUtils.join instead of String.join

jdragonbae:
This patch means fixing a crash when running browser on N OS device.
`String.join` is currently used
to make enable features into one string.
But `String.join` is supported only in java8 or higher.
So, use `TextUtils.join` instead of `String.join`.

Refs: #377


romandev:
Can we use `TextUtils.join` instead of `Strings.join` or `StringBuilder`?

Issue #378, opened 2020-04-20
A crash when calling hasSwitch('DnsOverHttps') in debug build


jdragonbae:
Cause of crash in this line[1]. 
`Check failed: ToLowerASCII(switch_string) == switch_string (dnsoverhttps vs. DnsOverHttps)'`
It seems that only lowercase characters are allowed as parameters.

[1] https://source.chromium.org/chromium/chromium/src/+/master:base/command_line.cc;l=300

PR #381, opened 2020-04-20
Change how to check the DnsOverHttps flag

jdragonbae:
This patch means changing the way to check the `DnsOverHttps` flag.
In BananaCommandLineInitializer, `enable-features` switch would be set
with the `DnsOverHttps`. So, by calling `getSwitchValue("enable-features")`,
we can check whether the `DnsOverHttps` flag is on.

Refs: #378


upstream/tb_net_android_java_src_org_chromium_net_DnsStatus.java.patch
```diff
@@ -7,7 +7,7 @@ diff --git tb/net/android/java/src/org/chromium/net/DnsStatus.java tb_diff/net/a
      public boolean getPrivateDnsActive() {
 -        return mPrivateDnsActive;
 +        return mPrivateDnsActive
-+                || org.chromium.base.CommandLine.getInstance().hasSwitch("DnsOverHttps");
++                || org.chromium.base.CommandLine.getInstance().hasSwitch("dnsoverhttps");
```
romandev, line 10: BTW, Does this really work?
I'm not sure if this is really working.. because `--dnsoverhttps` flag is not defined.
We probably should check feature flag. (not switch flag)

jdragonbae: Thanks for the detail.
You're right. It seems doesn't work. 
I try to check the ChromeFeatureList[1]. But, it's located in '/chrome'.
So, it can not be referenced in '/net'.
Is there another way to check the feature flag?

upstream/tb_net_android_java_src_org_chromium_net_DnsStatus.java.patch
```diff
@@ -1,13 +1,15 @@
...
+@@ -39,11 +39,14 @@ public class DnsStatus {
  
      @CalledByNative
      public boolean getPrivateDnsActive() {
 -        return mPrivateDnsActive;
 +        return mPrivateDnsActive
-+                || org.chromium.base.CommandLine.getInstance().hasSwitch("DnsOverHttps");
++                || org.chromium.base.CommandLine.getInstance()
++                           .getSwitchValue("enable-features")
++                           .contains("DnsOverHttps");
```
jdragonbae, line 12: @romandev I'm not sure this is the right way. PTAL

codeimpl: lgtm by @romandev

PR #568, opened 2020-08-19
Make BrowserLock start/stop no need to restart

knj0426:
When enable/disable BrowserLock, if user cancel restart application,
BrowserLock is not applied properly because TripleBananaApplication is
not called when next restart.

Signed-off-by: Nojeong Gwak <knj0426@gmail.com>


romandev (commented):
Thank you for this patch.

Your patch looks good to me overall but it has two commits.
If possible, we prefer to put one commit into one PR.
Could you please separate the dark mode related commit from this PR?

romandev:
Also, we don't use "Signed-off" in commit description.

src/org/triple/banana/settings/ExtensionFeatures.java
```diff
@@ -75,6 +75,7 @@ public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
                 (SwitchPreferenceCompat) findPreference(FeatureName.DARK_MODE);
         darkMode.setChecked(DarkModeController.get().isDarkModeOn());
         darkMode.setOnPreferenceChangeListener((preference, newValue) -> {
+            BrowserLock.setExceptional(true);
```
jdragonbae, line 78: Does this mean impossible on dark mode temporary?

knj0426: Yes. When toggle dark mode, application is restarted forcely. So, ApplicationStatus is changed background and foreground in a very short time. ApplicationApplicationTracker cannot avoid this behavior, so I though it should be handled as exceptional case.

knj0426:
> Thank you for this patch.
> 
> Your patch looks good to me overall but it has two commits.
> If possible, we prefer to put one commit into one PR.
> Could you please separate the dark mode related commit from this PR?

knj0426:
> Thank you for this patch.
> 
> Your patch looks good to me overall but it has two commits.
> If possible, we prefer to put one commit into one PR.
> Could you please separate the dark mode related commit from this PR?

Ok. I will do it

Issue #558, opened 2020-08-18
Remote control view doesn't disappear.


codeimpl:
The remote control view does not disappear when the back button is clicked during video playback(Check below sample).

https://www.thetvnamu.site/video/123174/정도.html

Issue #556, opened 2020-08-18
Video doesn't change to fullscreen.


codeimpl:
When the fullscreen button is clicked, the video does not change to fullscreen(Check below sample).

https://www.thetvnamu.site/watch_video/125077/%EB%AA%A8%EB%B2%94%ED%98%95%EC%82%AC-13%ED%9A%8C.html

PR #570, opened 2020-08-20
Fix the fullscreen related issues for RemoteControl service

romandev:
This patch fixes the fullscreen related issues for RemoteControl service
as follows:
  - Fix a bug that the RemoteControl view isn't shown on some websites
  - Fix a bug that the RemoteControl view isn't disappear after
    fullscreen mode is exited

Refs: #556, #558


jdragonbae:
LGTM!

codeimpl (approved):
lgtm

Issue #562, opened 2020-08-18
RemoteControl should support PIP mode on Android O+ OS


romandev:
RemoteControl should support PIP mode on Android O+ OS. Especially we should do the following things:
- Implement PIP mode button (using `BananaPictureInPictureControlller`)
- If user device's OS is lower than Android O, we can't support PIP mode because platform doesn't support. So, in that case, we should make a toast or hide the PIP button.

PR #571, opened 2020-08-20
 Support pip mode on remote control service

jdragonbae:
This patch means supporting pip mode on remote control service.
Picture-in-picture (PIP) mode only allowed on android 8.0 (API level 26) or
higher. The pip button only visible in supported versions. If pip mode is not
supported even though android 8.0 or higher, show the toast message.

Refs: #562


res/layout/remote_control_view.xml
```diff
@@ -48,7 +48,8 @@
                 style="@style/BananaMediaButtonTheme"
                 android:layout_marginRight="@dimen/margin_option_button"
                 android:layout_toLeftOf="@id/rotate_button"
-                android:src="@drawable/ic_pip_32dp" />
+                android:src="@drawable/ic_pip_32dp"
```
romandev, line 51: I think it's better to make the PIP button show in case of Android O+ programatically.

jdragonbae: Ack.

Issue #489, opened 2020-07-13
Implement video extensions


romandev:
We should support the following extension features during playing videos.
  - Lock (not allow any inputs during play)
  - Rotate video viewport
  - Full screen
  - PIP (if possible)

codeimpl:
1. lock 구현 해야함 (imageView toggle)
2. mute 구현 (imageView toggle)
3. seekbar 구현 (currentTime, duration)
4. play & pause (imageView toggle)
5. fast forward, fast rewi. nd 구현

PR #587, opened 2020-08-24
Make the sizes of the icons on remote control view smaller

romandev:
This patch makes the sizes of the icons on remote control view smaller.
(from 32dp to 24dp)

Refs: #489


jdragonbae (approved):
lgtm

Issue #565, opened 2020-08-18
Block background of BrowserLock screen


knj0426:
Background should not be visible when BrowserLock screen is on

PR #588, opened 2020-08-24
Block background when authencicate BrowserLock(#565)

knj0426:
When try to authenticate BrowserLock, backgrund should not be shown.

Refs #565


src/org/triple/banana/authentication/ActivityBasedBackend.java
```diff
@@ -5,18 +5,28 @@
...
 package org.triple.banana.authentication;
 
+import android.os.Bundle;
+
 import org.triple.banana.authentication.Authenticator.Callback;
 import org.triple.banana.base.InterActivity;
 
 class ActivityBasedBackend implements Backend {
     Class<? extends BaseActivity> mActivityClass;
+    private Bundle mExtras = null;
```
jdragonbae, line 15: AFAIK, this default value would be already null?

src/org/triple/banana/authentication/ActivityBasedBackend.java
```diff
@@ -5,18 +5,28 @@
...
+        if (mExtras == null) {
+            mExtras = new Bundle();
+        }
+        mExtras.putBoolean("BACKGROUND", translucent);
+    }
+
     @Override
     public void authenticate(Callback callback) {
-        InterActivity.start(mActivityClass, new Object(), callback::onResult);
+        InterActivity.start(mActivityClass, new Object(), callback::onResult, mExtras);
```
romandev, line 30: `InterActivity` provides a custom parameter and a custom callback.
You could simply pass the boolean value instead of `new Object()`.

src/org/triple/banana/authentication/BaseActivity.java
```diff
@@ -5,12 +5,23 @@
...
+import android.os.Bundle;
+
+import org.triple.banana.R;
 import org.triple.banana.base.InterActivity;
 
 /**
  * BaseActivity
  */
 public class BaseActivity extends InterActivity<Object, Boolean> {
+    @Override
```
romandev, line 17: `public class BaseActivity extends InterActivity<Boolean /* background */, Boolean /* result */>`

src/org/triple/banana/authentication/BaseActivity.java
```diff
@@ -5,12 +5,23 @@
...
+    public void onCreate(Bundle savedInstanceState) {
```
romandev, line 18: Let's remove this change

src/org/triple/banana/base/InterActivity.java
```diff
@@ -27,7 +27,7 @@
     private Callback<Response> mCallback;
 
     public static <R> void start(Class<? extends InterActivity> activityClass, Object requestData,
-            Callback<R> callback) {
+            Callback<R> callback, Bundle bundle) {
```
romandev, line 30: Let's remove this change

src/org/triple/banana/base/InterActivity.java
```diff
@@ -36,6 +36,9 @@
         Context context = BananaApplicationUtils.get().getApplicationContext();
         Intent intent = new Intent(context, activityClass);
         intent.putExtra("__request_id", sRequestId);
+        if (bundle != null) {
```
romandev, line 39: Let's remove this change

src/org/triple/banana/authentication/ActivityBasedBackend.java
```diff
@@ -5,18 +5,28 @@
...
 
 class ActivityBasedBackend implements Backend {
     Class<? extends BaseActivity> mActivityClass;
+    private Bundle mExtras;
 
     ActivityBasedBackend(Class<? extends BaseActivity> activityClass) {
         mActivityClass = activityClass;
     }
 
+    public void setBackground(boolean translucent) {
```
romandev, line 21: Let's remove this

src/org/triple/banana/authentication/ActivityBasedBackend.java
```diff
@@ -5,18 +5,28 @@
...
+        if (mExtras == null) {
+            mExtras = new Bundle();
+        }
+        mExtras.putBoolean("BACKGROUND", translucent);
+    }
+
     @Override
     public void authenticate(Callback callback) {
```
romandev, line 29: This method should take `authenticate(boolean, Callback)`

src/org/triple/banana/authentication/ActivityBasedBackend.java
knj0426: Thank you. I will update it today :)

src/org/triple/banana/authentication/BaseActivity.java
knj0426: But it looks like needs to call setContentView.
IMHO, InterActivity is general activity, so it is not good to check background.
So, I think we should check and set background here.
How do you think?

src/org/triple/banana/authentication/ActivityBasedBackend.java
```diff
@@ -17,6 +17,11 @@
 
     @Override
     public void authenticate(Callback callback) {
-        InterActivity.start(mActivityClass, new Object(), callback::onResult);
+        InterActivity.start(mActivityClass, Boolean.FALSE, callback::onResult);
```
romandev, line 20: `authenticate(false, callback);`

src/org/triple/banana/authentication/BaseActivity.java
romandev: You're correct!

src/org/triple/banana/authentication/BaseActivity.java
```diff
@@ -5,12 +5,23 @@
...
 /**
  * BaseActivity
  */
-public class BaseActivity extends InterActivity<Object, Boolean> {
+public class BaseActivity extends InterActivity<Boolean, Boolean> {
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        Boolean isBackground = getRequestData();
+        if (isBackground != null && isBackground.equals(Boolean.TRUE)) {
```
romandev, line 21: We can simplify 20 ~ 23 lines as follows:
```
if (Boolean.TRUE.equals(getRequestData()) {
  setContentView(R.layout.browser_lock_background);
}
```

romandev (commented):
lgtm with nits

romandev (approved):
lgtm

Issue #546, opened 2020-08-17
Authenticator.authenticate calls Callback twice.


knj0426:
When authentication done, Authenticator calls Callback twice.
1. In case of success : true, false
2. In case of failure : false, false

PR #590, opened 2020-08-24
Fix a bug that the authentication callback is duplicately called twice

romandev:
The bug is caused that current implementation handles the authentication
callback in onPause(). In the case of BiometricPrompt, it is
automatically terminated when losting a focus on its UI, so there is no
need to finalize the callback in onPause().

Refs: #546


jdragonbae (approved):
LGTM

Issue #578, opened 2020-08-23
Make the authentication module use androidx.biometric.BiometricPrompt


romandev:
Currently, our authentication module uses each different backend depending on the platform as follows:
  - Android P+: `BiometricPrompt` backend
  - Android M+: `FingerprintManager` backend
  - Android L: `Keyguard` backend

By the way, there are some bugs that the `FingerprintManager` backend is not working well on Android O (or other platforms). Also, we don't want to split our code on each platforms. So, it's better to use `androidx.biometric.BiometricPrompt` than fixing the `FingerprintManager` backend.

PR #591, opened 2020-08-24
Use Keygaurd as a fallback when the biometric sensor is unavailable

romandev:
In the current implementation, the BiometricPrompt is terminated when
getting ERROR_CANCELED even if BiometricManager.canAuthenticate() was
success. It is because the biometric sensor is unavilable for some
reasons. Therefore, this patch provides a fallback using Keyguard
authentication if possible.

Refs: #578


codeimpl:
LGTM

codeimpl (approved):
LGTM

Issue #554, opened 2020-08-18
MediaCommandProcessor is not working if media session was gone


romandev:
This is a regression after #536. We should find a way to implement this without `MediaSession`.

Issue #535, opened 2020-08-13
Remote control's play() button is not working if autoplay is disabled


romandev:
Currently, if autoplay feature is disabled, remote control view's play() button is not working.
It is caused because the `MediaCommandProcessor` can't consume user gesture because the remote control view is overlayed.

PR #592, opened 2020-08-25
Use runMediaCommand instead of BananaMediaCommandProcessor

romandev:
`BananaMediaCommandProcessor` is not working if media session was gone.
This is a regression after #536. So, this patch partially reverts the
patch and leave a `FIXME` comment.

Refs: #535, #554


jdragonbae:
LGTM

jdragonbae (approved):

src/org/triple/banana/media/MediaCommandProcessor.java
```diff
@@ -32,19 +32,21 @@ private void runMediaCommand(String mediaCommand) {
...
     }
 
     public void pause() {
-        getCommandProcessor().pause();
+        runMediaCommand("pause()");
     }
 
-    public void setPosition(long ms) {
-        getCommandProcessor().setPosition(ms);
+    public void setPosition(float value) {
```
jdragonbae, line 44: Maybe, we will use `double` type for the position value. I wonder `float` type is okay.

romandev: This class is `MediaCommandProcessor`. It means that this class is not related to `ViewModel`. It should be managed by ViewModel itself.

PR #593, opened 2020-08-25
Change MediaController according to MediaCommandProcessor

jdragonbae:
This patch means changing MediaController according to
MediaCommandProcessor.

Refs: #592


romandev (approved):
lgtm

Issue #596, opened 2020-08-25
ViewModelData is not reset when restarting RemoteControlService


romandev:
There is a problem with the previous value being kept after restarting remote control service. It also makes the brightness value crazy.

PR #598, opened 2020-08-25
Introduce ViewModel.reset() to reset the model data

romandev:
This patch fixes the following problems:
  - The previous model data is kept after restarting remote control
    service
  - The brightness value is sometimes crazy
  - GestureDetector doesn't work after restarting remote control service

Refs: #596, #597


jdragonbae (approved):
LGTM!

src/org/triple/banana/remote_control/RemoteControlViewModel.java
```diff
@@ -35,8 +38,15 @@ boolean getIsLocked() {
     }
 
     static class Data extends ReadonlyData implements Cloneable {
+        void reset() {
+            setBrightness(BrightnessUtil.SYSTEM_DEFAULT);
+            setVolume(AudioUtil.getMediaVolume());
+            setControlsVisibility(true);
+            setIsLocked(false);
```
jdragonbae, line 45: Could you please reset for `position`?

romandev: Done

Issue #557, opened 2020-08-18
Fix sensitivity for gesture detector


jdragonbae:
We need to fix a sensitivity for gesture detector via threshold.

PR #599, opened 2020-08-25
Fix sensitivity for gesture detection

romandev:
This patch makes the gesture detector pass the percentage value for
distance(dx or dy) to the callback.

Refs: #557


jdragonbae (approved):
lgtm

Issue #553, opened 2020-08-18
Implement forward and rewind when the user double-clicks the screen


jdragonbae:
This is one of the features of remote control service. 
Example) Forward 10 seconds if user clicks twice.

PR #600, opened 2020-08-25
Implement double tap gestures for forward/backward operations

romandev:
This patch implements double tap gestures for forward/backward
operations.

Refs: #553


jdragonbae (approved):
lgtm

Issue #585, opened 2020-08-23
Implement UI when volume/brightness is changed by gesture detector


romandev:
Currently, we have no UI when volume/brightness is changed by gesture detector. It isn't too user friendly. So, we should implement the related UI when the value is changed.

PR #602, opened 2020-08-26
Make the gesture detector trigger the only one gesture event at a time

romandev:
This patch makes the gesture detector trigger the only one gesture event
at a time by storing/checking the current detected gesture. To simplify
this problem, we don't add any new `finish` events. Instead we just
assume the finished state if the changed value is 0.

Refs: #585
Co-authored-by: Jaeyong Bae <jdragon.bae@gmail.com>


jdragonbae (approved):
lgtm

Issue #604, opened 2020-08-26
Implement additional media events on MediaEventDispatcher


romandev:
We need the following events:
  - onPlay
  - onPause
  - onWait?
  - onTimeUpdate
  - Etc.

PR #614, opened 2020-08-27
Implement additional media events on MediaEventDispatcher

romandev:
This patch implements the following events:
  - OnPlayStateChanged(): This is called when the play state on media
    element is changed. The state can be one of three states as follows:
      - PLAYING: The media element is playing.
      - PAUSED: The media element is paused.
      - WAITING: The media element is playing but there is no enough
        data to play.
  - OnTimeUpdate(): This is called when the media element's duration or
    current time is updated.

Refs: #604


jdragonbae (approved):
lgtm

Issue #609, opened 2020-08-26
Display forward/backward state UI when the forward/backward is called


romandev:
Currently, we have no UI when the forward/backward is called on RemoteControlView. Especially, if the forward/backward is called from double tap gesture, it is difficult for users to know the changes. Therefore, we should display something when it happens.

PR #616, opened 2020-08-27
Add ripple effect when a user double-clicks

jdragonbae:
This patch means adding ripple effect when a user double-clicks screen.
It works the same as forward and backward operations and makes the
ripple effect visible.

Refs: #609
Co-authored-by: Byoungkwon Ko <codeimpl@gmail.com>


src/org/triple/banana/remote_control/RemoteControlService.java
```diff
@@ -190,14 +190,20 @@ public void onControlsStateChanged() {
     @Override
     public void onBackward() {
         if (mViewModel.getData().getIsLocked()) return;
-
         mMediaController.setRelativePosition(-10.0f);
+        mViewModel.getEditor().setEffect(RemoteControlViewModel.Effect.BACKWARD);
```
romandev, line 194: It's not bad but I think this approach has some disadvantages:
  - This approach causes invalidations twice for entire view. Because we have no dirty bits, this might be too heavy.
  - Currently, it works anyway but if the `showEffect()` or other operations are asynchronous, it might not work well.

Hence, I'd like to suggest `setXXX` and `takeXXX` method pair.

jdragonbae: Thanks for the review. I'm not sure whether the way is right way. PTAL.

src/org/triple/banana/remote_control/RemoteControlService.java
```diff
@@ -190,14 +190,18 @@ public void onControlsStateChanged() {
...
+        mViewModel.commit();
+        mViewModel.getEditor().setEffect(RemoteControlViewModel.Effect.UNKNOWN);
```
romandev, line 196: You don't need to set this value here. The `takeXXX` means the value is moved to other object.

src/org/triple/banana/remote_control/RemoteControlViewModel.java
```diff
@@ -16,6 +16,8 @@
 class RemoteControlViewModel {
     private static final String TAG = "RemoteControlViewModel";
 
+    enum Effect { UNKNOWN, FORWARD, BACKWARD }
```
romandev, line 19: `NONE, FORWARD, BACKWARD`?

src/org/triple/banana/remote_control/RemoteControlViewModel.java
```diff
@@ -26,6 +28,7 @@
         protected boolean mVolumeControlVisibility;
         protected boolean mBrightnessControlVisibility;
         protected MediaPlayState mPlayState;
+        protected Effect mEffect;
```
romandev, line 31: This should be initial value because this is not basic type.

src/org/triple/banana/remote_control/RemoteControlViewImpl.java
```diff
@@ -151,6 +152,24 @@ private void setPlayState(MediaPlayState state) {
         }
     }
 
+    private void showEffect(RemoteControlViewModel.Effect effect) {
+        View view = null;
```
romandev, line 156:
```
View view;
switch (effect) {
    case BACKWARD:
        view = mDialog.findViewById(R.id.backward_ripple);
        break;
    case FORWARD:
        view = mDialog.findViewById(R.id.forward_ripple);
        break;
    case NONE:
    default:
        view = null;
}

...
```

romandev:
Any update?

src/org/triple/banana/remote_control/RemoteControlViewImpl.java
jdragonbae: Done.

src/org/triple/banana/remote_control/RemoteControlViewModel.java
jdragonbae: Thanks. Done.

src/org/triple/banana/remote_control/RemoteControlViewModel.java
jdragonbae: Done.

src/org/triple/banana/remote_control/RemoteControlService.java
```diff
@@ -190,14 +190,14 @@ public void onControlsStateChanged() {
     @Override
     public void onBackward() {
         if (mViewModel.getData().getIsLocked()) return;
-
         mMediaController.setRelativePosition(-10.0f);
+        mViewModel.setOneTimeValue(RemoteControlViewModel.Value.BACKWARD_EFFECT);
```
romandev, line 194: Let's simply call `mView.showEffect(...);` here.
I think the view model is not good for this.

src/org/triple/banana/remote_control/RemoteControlViewModel.java
```diff
@@ -16,6 +16,8 @@
 class RemoteControlViewModel {
     private static final String TAG = "RemoteControlViewModel";
 
+    enum Value { NONE, FORWARD_EFFECT, BACKWARD_EFFECT }
```
romandev, line 19: Let's move this to View
and should be `enum Effect { NONE, FORWARD, BACKWARD }`

src/org/triple/banana/remote_control/RemoteControlViewModel.java
```diff
@@ -26,6 +28,7 @@
         protected boolean mVolumeControlVisibility;
         protected boolean mBrightnessControlVisibility;
         protected MediaPlayState mPlayState = MediaPlayState.PAUSED;
+        protected Value mValue = Value.NONE;
```
romandev, line 31: Let's remove this

src/org/triple/banana/remote_control/RemoteControlViewImpl.java
```diff
@@ -151,6 +152,26 @@ private void setPlayState(MediaPlayState state) {
         }
     }
 
+    private void showEffect(RemoteControlViewModel.Value value) {
```
romandev, line 155: Let's make this public.

src/org/triple/banana/remote_control/RemoteControlService.java
```diff
@@ -190,14 +190,14 @@ public void onControlsStateChanged() {
...
     }
 
     @Override
     public void onForward() {
         if (mViewModel.getData().getIsLocked()) return;
-
         mMediaController.setRelativePosition(10.0f);
+        mViewModel.setOneTimeValue(RemoteControlViewModel.Value.FORWARD_EFFECT);
```
romandev, line 201: We can just call `mView.showEffect(...);` here.

src/org/triple/banana/remote_control/RemoteControlViewImpl.java
```diff
@@ -67,6 +67,7 @@ public void onUpdate(RemoteControlViewModel.ReadonlyData data) {
         showControls(data.getControlsVisibility(), data.getIsLocked(), data.getIsVolumeMuted());
         setPosition(data.getPosition());
         setPlayState(data.getPlayState());
+        showEffect(data.takeValue());
```
romandev, line 70: Let's remove this

src/org/triple/banana/remote_control/RemoteControlViewModel.java
jdragonbae: Done.

src/org/triple/banana/remote_control/RemoteControlViewModel.java
jdragonbae: Done.

src/org/triple/banana/remote_control/RemoteControlViewImpl.java
jdragonbae: Done.

src/org/triple/banana/remote_control/RemoteControlViewImpl.java
jdragonbae: Done.

src/org/triple/banana/remote_control/RemoteControlService.java
jdragonbae: SGTM. Done.

src/org/triple/banana/remote_control/RemoteControlService.java
jdragonbae: SGTM.

src/org/triple/banana/remote_control/RemoteControlView.java
```diff
@@ -10,6 +10,8 @@
 import androidx.annotation.NonNull;
 
 interface RemoteControlView {
+    enum Effect { NONE, FORWARD, BACKWARD }
+
```
romandev, line 14: `showEffect` method is missing here.

jdragonbae: Done.

romandev (approved):
lgtm

PR #615, opened 2020-08-27
Make the play/pause buttons toggled

jdragonbae:
This patch means making the play/pause buttons toggled by using state.
The state in view model will be in 3 such as PLAY, PAUSE, WAITING. When
waiting state, the progress bar will be show and hide the middle
control.

Refs: #584
Co-authored-by: Byoungkwon Ko <codeimpl@gmail.com>
Co-authored-by: Jinho Bang <zino@chromium.org>

PR #617, opened 2020-08-27
Fix a crash when entering remote control service

romandev:
This is a regression since #615. We should initialize the variable.

Refs: none


codeimpl (approved):
LGTM

jdragonbae (approved):
lgtm

Issue #612, opened 2020-08-26
Implement metadata in model and view


romandev:
Currently, even if playing the video on RemoteControl service, the metadata (current time, duration) is not updated yet. We should implement them into model and update them on View (such as seekbar slide and text)

PR #618, opened 2020-08-27
Implement metadata update and seeking on remote control

romandev:
Currently, even if playing the video on RemoteControl service, the
metadata(current time, duration) is not updated yet. We should implement
them into model and update them on View(such as seekbar and label text).

This patch includes the following things:
  - Time update
  - Seeking on SeekBar
  - Rename setPosition with setCurrentTime (in remote_control module)

Refs: #612


src/org/triple/banana/remote_control/RemoteControlService.java
```diff
@@ -159,13 +166,12 @@ public void onBrightnessChanged(float value) {
...
-        float position = mViewModel.getEditor().getPosition();
+    public void onCurrentTimeChanged(double diffTime) {
+        double newTime = mViewModel.getEditor().getCurrentTime() + diffTime;
         mViewModel.getEditor().setControlsVisibility(true);
-        mViewModel.getEditor().setPosition(position + value);
+        mViewModel.getEditor().setCurrentTime(newTime);
         mViewModel.commit();
-        // FIXME(#589): Need to fix seekTo operation of MediaSession
-        // mMediaController.setPosition(ms);
+        mMediaController.setPosition((float) (newTime));
```
jdragonbae, line 174: Can we call setPosition when the diff time is 0.0?

src/org/triple/banana/remote_control/RemoteControlService.java
```diff
@@ -159,13 +166,12 @@ public void onBrightnessChanged(float value) {
     }
 
     @Override
-    public void onPositionChanged(float value) {
-        float position = mViewModel.getEditor().getPosition();
+    public void onCurrentTimeChanged(double diffTime) {
+        double newTime = mViewModel.getEditor().getCurrentTime() + diffTime;
```
jdragonbae, line 170: Do we need check isLocked? 
`if (mViewModel.getData().getIsLocked()) return;`

src/org/triple/banana/remote_control/RemoteControlService.java
romandev: I introduced `onPositionChangeFinish()` and used it. Done.

src/org/triple/banana/remote_control/RemoteControlService.java
romandev: Done.

jdragonbae (approved):
lgtm

PR #621, opened 2020-08-27
Fix unexpected BrowserLock with KeyguardActivity(#565)

knj0426:
In case of backend with KeyguardActivity, it sent application to background.
So we should apply exceptional condition for this case.
But activity and application status changing condition is different with
setting dark mode.
So I separate it into two case. 1) Application restart 2) Keyguard case


romandev:
@knj0426 Any update?

knj0426:
> @knj0426 Any update?

I will upload session concept patch instead of this.

romandev:
Closed by #653

PR #622, opened 2020-08-27
Fix a bug the remote control view is not shown after PIP mode

jdragonbae:
This patch fixes a bug the remote control view is not shown after PIP
mode. Instead of using pip controller, it works same as home button
clicked.

Refs: #620


jdragonbae:
After this patch, there is an issue where the navigation bar does not disappear.

romandev:
Did you try the approach to get the PIP controller member in ChromeActivity?

jdragonbae:
> Did you try the approach to get the PIP controller member in ChromeActivity?

Yes. It seems to work as same.

upstream/tb_chrome_android_java_src_org_chromium_chrome_browser_ChromeActivity.java.patch
```diff
@@ -13,7 +13,21 @@ diff --git tb/chrome/android/java/src/org/chromium/chrome/browser/ChromeActivity
...
              public void onCrash(Tab tab) {
                  postDeferredStartupIfNeeded();
-@@ -2024,9 +2029,13 @@ public abstract class ChromeActivity<C extends ChromeActivityComponent>
+@@ -924,6 +929,13 @@ public abstract class ChromeActivity<C extends ChromeActivityComponent>
+         getManualFillingComponent().onResume();
+     }
+ 
++    public PictureInPictureController getPipController() {
++        if (mPictureInPictureController == null) {
++            mPictureInPictureController = new PictureInPictureController();
```
romandev, line 22: Do we have to create a new instance?

jdragonbae:
I changed it to get the PIP controller in ChromeActivity.

banana_cake/java/org/banana/cake/CakePipController.java
```diff
@@ -25,7 +25,14 @@
             mPipController = null;
             return;
         }
-        mPipController = new PictureInPictureController();
+
+        ChromeActivity activity = getLastTrackedFocusedChromeActivity();
+        if (activity == null) {
+            mPipController = null;
```
romandev, line 31: We don't need to keep this because this causes a memory leak.

jdragonbae: Ack.

upstream/tb_chrome_android_java_src_org_chromium_chrome_browser_ChromeActivity.java.patch
jdragonbae: We can just return `mPictureInPictureController`.

romandev: Make sense!

romandev (approved):
LGTM
Just query: Does your patch fix the issue that the home button is shown when back to the fullscreen?

jdragonbae:
> LGTM
> Just query: Does your patch fix the issue that the home button is shown when back to the fullscreen?

I'm trying to fix the issue you said!

Issue #626, opened 2020-08-30
A crash occurs on Naver TV with remote control


romandev:
**Reproducible steps**
1. Run Banana Browser
2. Access https://m.tv.naver.com/v/15503746
3. Click the fullscreen button on the player
4. A crash occurs

**Callstack**

```
[...]

PR #627, opened 2020-08-30
Fix a crash on Naver TV with remote control service

romandev:
This is because the remote control view's `show()` is called after the
parent activity is already gone. To fix the problem, this patch makes
the `show()` method take `Activity` instead of `Context` and add
`isFinishing()` check.

Refs: #626


jdragonbae (approved):
lgtm

PR #628, opened 2020-08-30
Make BrowserLock class as singleton(#565)

knj0426:
All current methods of BrowserLock are static. There is no need to be static.
Make BrowserLock as singleton and remove static from methods and members.

Refs: #565


src/org/triple/banana/lock/BrowserLock.java
```diff
@@ -11,9 +11,19 @@
...
+
+    /**
+     * Get BrowserLock singleton instance.
+     * @return Instance of BrowserLock
+     */
+    public static BrowserLock getInstance() {
+        return LazyHolder.INSTANCE;
+    }
+
+    private static class LazyHolder { private static BrowserLock INSTANCE = new BrowserLock(); }
```
jdragonbae, line 24: nit: `private static final  BrowserLock INSTANCE`

jdragonbae:
lgtm

src/org/triple/banana/lock/BrowserLock.java
knj0426: Thank you. it's done!

PR #629, opened 2020-08-30
Make BrowserLock class as singleton

knj0426:
All current methods of BrowserLock are static. There is no need to be static.
Make BrowserLock as singleton and remove static from methods and members.

Refs: #565


romandev (approved):
lgtm

PR #630, opened 2020-08-30
Fix a bug navbar is shown when back to fullscreen

jdragonbae:
This patch fixes a bug navbar is shown when back to fullscreen after PIP
mode. When back to fullscreen after PIP mode, the system UI would be
changed. So, we could listen to that and hide system UI like navbar.

Refs: #624


romandev (approved):
lgtm

Issue #608, opened 2020-08-26
Change the lock state icon on RemoteControlView


romandev:
Currently, it is difficult for user to know the state of the lock on RemoteControlView.
We need to change the lock icon to understand easily.

PR #632, opened 2020-08-31
Change the lock state icon on RemoteControlView

codeimpl:
To increase visibility, this patch changes the lock state icon.

Refs: #608


romandev (approved):
LGTM

PR #634, opened 2020-08-31
Fix a bug that `update_chromium` command couldn't fetch the remote tags

romandev:
There is a bug that the `update_chromium` command couldn't fetch the
remote tags sometimes. It fails because it doesn't write a local
reference. So, this patch makes the local reference create by force.

Refs: #633


jdragonbae (approved):
lgtm

PR #635, opened 2020-08-31
Show navbar when the controls are visible

jdragonbae:
This patch enables showing the navbar when the controls are visible. It
would show/hide with controls.

Refs: #631


src/org/triple/banana/remote_control/RemoteControlViewImpl.java
```diff
@@ -35,6 +36,8 @@
     private @Nullable SeekBar mTimeSeekBar;
     private @Nullable Dialog mDialog;
     private @Nullable RemoteControlLayout mMainView;
+    private @Nullable Activity mParentActivity;
```
romandev, line 39: You could use getContext() without storing this.

src/org/triple/banana/remote_control/RemoteControlViewImpl.java
```diff
@@ -54,16 +57,23 @@ public void show(@NonNull Activity parentActivity) {
...
+        }
     }
 
     @Override
     public void dismiss() {
         if (mDialog == null) return;
-        mDialog.getWindow().getDecorView().setOnSystemUiVisibilityChangeListener(null);
         mDialog.dismiss();
         mRemoteControlGestureDetector.stopDetection();
+        if (mParentActivity != null && mParentActivity.getWindow() != null) {
```
romandev, line 72: You could use getContext()

src/org/triple/banana/remote_control/RemoteControlViewImpl.java
```diff
@@ -35,6 +36,8 @@
     private @Nullable SeekBar mTimeSeekBar;
     private @Nullable Dialog mDialog;
     private @Nullable RemoteControlLayout mMainView;
+    private @Nullable Activity mParentActivity;
+    private boolean mShowSystemUI;
```
romandev, line 40: Do we really need this flag? We can not easily sync this value.

src/org/triple/banana/remote_control/RemoteControlViewImpl.java
jdragonbae: Ack.

src/org/triple/banana/remote_control/RemoteControlViewImpl.java
jdragonbae: Done.

src/org/triple/banana/remote_control/RemoteControlViewImpl.java
jdragonbae: Ack.

src/org/triple/banana/remote_control/RemoteControlViewImpl.java
```diff
@@ -54,16 +55,23 @@ public void show(@NonNull Activity parentActivity) {
...
 
     @Override
     public void dismiss() {
         if (mDialog == null) return;
-        mDialog.getWindow().getDecorView().setOnSystemUiVisibilityChangeListener(null);
         mDialog.dismiss();
         mRemoteControlGestureDetector.stopDetection();
+        BananaTab tab = org.banana.cake.interfaces.BananaTabManager.get().getActivityTab();
+        if (tab == null) return;
+        Activity activity = (Activity) tab.getContext();
```
romandev, line 71: The null check is unnecessary because the `getContext()` is always non-null.
https://developer.android.com/reference/android/app/Dialog?hl=en#getContext()

jdragonbae: Thanks. Done.

romandev: Should be `mDialog.getContext()` instead of `tab.getContext()`

romandev (approved):
LGTM

PR #638, opened 2020-08-31
Apply session concept for BrowserLock(#565)

knj0426:
It is bothering user that showing BrowserLock everytime while using
browser.
To improve usability, apply session concept. If user authenticate once,
no need to authenticate again while session is alive(60sec).

Refs: #565


src/org/triple/banana/lock/BrowserLock.java
```diff
@@ -41,11 +59,22 @@ public void start() {
...
 
     public void setExceptional(boolean isException) {
         isExceptional = isException;
     }
+
+    private void startSession() {
+        mSessionTimer = new Timer();
+        mSessionTimer.schedule(new TimerTask() {
+            @Override
+            public void run() {
```
jdragonbae, line 75: Can we use lambda for this?

knj0426: Only interface which has one method. TimerTask is not a interface.

jdragonbae: Ack!

src/org/triple/banana/lock/BrowserLock.java
```diff
@@ -41,11 +59,22 @@ public void start() {
...
         ApplicationStatusTracker.getInstance().removeListener(mListener);
     }
 
     public void setExceptional(boolean isException) {
         isExceptional = isException;
     }
+
+    private void startSession() {
+        mSessionTimer = new Timer();
+        mSessionTimer.schedule(new TimerTask() {
```
romandev, line 73: In my humble opinion, this looks a little bit unsafe. Unlike our expectation, the `isAuthenticated` can be `true` forever. The `TimerTask` doesn't run for some reasons. It would be better to create a time token and check that the time token is valid when requesting an authentication. WDYT?

knj0426: I agree that. Can I consider timetoken as timestamp? Which is the better way to store timestamp? 1. just variable? 2. using DB?
1. it cannot maintain timestamp in case of reboot. so, after reboot, user should authenticate event timeout hasn't happened.
2. in case of using Date and Time as timestamp, If user change timezone or set time manually, it can cause exceptional case. 
in case of using CPU counter as timestamp, if user reboot device, this value is set to zero. and also can cause exceptional case.

knj0426: IMHO, it would better just using variable and reset if user reboot device.

Issue #559, opened 2020-08-18
VideoRemoteControl suddenly turns off without user gesture when buffering situation


romandev:
Sometimes VideoRemoteControl suddenly turns off when buffering while running VideoRemoteControl on some websites (e.g. YouTube)

**Reproducible steps**
1. Launch Banana Browser
2. Access https://youtube.com
3. Play any movie
4. Enter fullscreen mode
5. Exit fullscreen mode
6. Repeat 4-5 until buffering happens
7. When buffering is keep going, enter fullscreen mode and then wait to observe the issue

PR #639, opened 2020-09-01
Fix a bug that the the remote control is gone suddenly during waiting

romandev:
The problem is because `exitFullscreen()` from web site is not blocked.
During the playing the media on remote control service, we should block
the `exitFullscreen()` that is called on web site.

Refs: #559


jdragonbae (approved):
lgtm

Issue #619, opened 2020-08-27
The save password prompt is not shown on naver.com after login failure


romandev:
**Reproducible steps:**
1. Run Banana Browser
2. Access https://m.naver.com
3. Try to login with your ID and WRONG password
4. Retry to login with your ID and your correct password
5. After login, you can't see the save password prompt

**Expected behavior:**
The save password prompt should be shown

**Actual behavior:**
The save password prompt is NOT shown

romandev:
I made a patch to fix this issue on upstream.
https://chromium-review.googlesource.com/c/chromium/src/+/2379434

romandev:
Reopen for tracking upstream change. If the upstream change is merged, we should remove our side change.

PR #641, opened 2020-09-01
Fix a bug that the save password prompt isn't shown after login failure

romandev:
There is a bug that the save password prompt isn't shown after
login failure on naver.com. It is because the naver.com intentionally
reset the ID/password field before submmiting the login form.
So, this patch fixes the problem by getting the ID value from attribute
intead of non-attribute-value if the value already gotten is empty.

Refs: #619


jdragonbae (approved):
lgtm

PR #648, opened 2020-09-02
Add overflow checks in onPositionChanged()

romandev:
Currently, when onPositionChanged() is called by GestureDetector, there
is a problem that the limit value in seekbar might be overflowed. So,
this patch adds overflow checks in onPositionChanged() and revise the
values.

Refs: #643


jdragonbae (approved):
lgtm

codeimpl (approved):
LGTM

PR #651, opened 2020-09-02
Add a switch option on settings for media remote feature

romandev:
The feature option is enabled by default.

Refs: #567


jdragonbae (approved):
lgtm

PR #652, opened 2020-09-02
When video state is in a waiting state, playback smoothly

jdragonbae:
This patch makes playback smooth by calling the set position when the
video is in a waiting state. It is called 1 seconds after a waiting
state.

Refs: none


src/org/triple/banana/media_remote/MediaRemoteService.java
```diff
@@ -53,6 +56,11 @@ public void start() {
             public void onPlayStateChanged(MediaPlayState state) {
                 mViewModel.getEditor().setPlayState(state);
                 mViewModel.commit();
+                if (state == MediaPlayState.WAITING) {
```
romandev, line 59: 1. Let's reduce time from 2 seconds to 1 second.
2. We only should retry seeking when the previous play state was PLAYING and the current play state is WAITING now.
3. We should removeCallbacks whenver the state is chagned.
```
mHandler.removeCallbacks(mTaskToRetrySeeking);
if (mViewModel.getData().getPlayState() == MediaPlayState.PLAYING && state == MediaPlayState.WAITING) {
    mHandler.postDelayed(mTaskToRetrySeeking, 1000);
}
mViewModel.getEditor().setPlayState(state);
mViewModel.commit();
```

jdragonbae: Ack.

jdragonbae:
@romandev  Thanks for the review. PTAL

romandev (approved):
lgtm

PR #653, opened 2020-09-02
Apply session concept for BrowserLock

knj0426:
It is bothering user that showing BrowserLock everytime while using
browser.
To improve usability, apply session concept. If user authenticate once,
no need to authenticate again while session is alive(60sec).

Refs: #565


src/org/triple/banana/lock/BrowserLock.java
```diff
@@ -48,4 +59,13 @@ public void stop() {
     public void setExceptional(boolean isException) {
         isExceptional = isException;
     }
+
+    private void setTimestamp() {
+        authTimestamp = (long) (System.currentTimeMillis());
+    }
+
+    private boolean isSessionExpired() {
```
romandev, line 67: As you already know, this is unsafe but we can improve this later (e.g. using Android KeyStore).
Instead please file an issue and leave a FIXMEcomment as follows:
```
// FIXME(#xxx): This is insecure but for now blah blah..
```

src/org/triple/banana/lock/BrowserLock.java
```diff
@@ -48,4 +59,13 @@ public void stop() {
...
+        long currentTimestamp = (long) (System.currentTimeMillis());
+        return currentTimestamp - authTimestamp > SESSION_DURATION;
```
romandev, line 69: As I already mentioned above, this is basically unsafe. If the users adjust their system time to make the `currentTime` less than `authenticationTime`, this method is always `false` and authentication is ommitted. But even so, I think it could be safe than now by tightening the conditions as follows:
```
long currentTime = System.currentTimeMillis();
return currentTime < mLastAuthenticationTime || currentTime - mLastAuthenticationTime > SESSION_DURATION;
```

More thing, we can consider to listen to time change event via broadcast receiver. But it would better to make it in separated patch if you want.

src/org/triple/banana/lock/BrowserLock.java
```diff
@@ -25,12 +29,19 @@ public static BrowserLock getInstance() {
 
     private ApplicationStatusListener mListener = (lastActivity, status) -> {
         if (status == ApplicationStatus.FOREGROUND) {
+            if (!isSessionExpired()) {
+                return;
+            }
```
romandev, line 34: If the session is expired, then you should reset the `mLastAuthenticationTime` to null.

romandev (commented):
Patch looks good to me.
I left some comments.
(Please don't add (#xxx) in the commit title.)

Thank you.

src/org/triple/banana/lock/BrowserLock.java
knj0426: I thought even session is expired, if user canceled or failed authentication, we should check mLastAuthenticationTime for retry. As you said, even mLastAuthenticationTime is set to null or not action will be same. I just though no need to add more code. Did I miss something?

romandev: > I thought even session is expired, if user canceled or failed authentication, we should check mLastAuthenticationTime for retry.

There are a number of reasons this is unsafe, but the biggest problem is that the users can change their system time on device.

> As you said, even mLastAuthenticationTime is set to null or not action will be same. I just though no need to add more code. Did I miss something?

romandev: > I thought even session is expired, if user canceled or failed authentication, we should check mLastAuthenticationTime for retry. As you said, even mLastAuthenticationTime is set to null or not action will be same. I just though no need to add more code. Did I miss something?

There are a number of reasons this is unsafe, but the biggest problem is that the users can change their system time on device.

If the stored value is not reset, the value is kept persistently until authentication succeeded. Users can go back to the past by changing  the device time, and then bypass the authentication eventually.

If the stored value is reset, there is no values to compare, so even if the users goes back to the past via changing the device time, they should reauthenticate because the store value is not exist.

src/org/triple/banana/lock/BrowserLock.java
```diff
@@ -13,6 +13,10 @@
 public class BrowserLock {
     private boolean isExceptional = false;
 
+    // BrowserLock Session
+    private long authTimestamp;
```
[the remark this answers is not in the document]
knj0426: Got it!

src/org/triple/banana/lock/BrowserLock.java
```diff
@@ -48,4 +59,13 @@ public void stop() {
     public void setExceptional(boolean isException) {
         isExceptional = isException;
     }
+
+    private void setTimestamp() {
```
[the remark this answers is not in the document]
knj0426: got it!

src/org/triple/banana/lock/BrowserLock.java
knj0426: I agree that. I will make fix it on another patch

src/org/triple/banana/lock/BrowserLock.java
knj0426: I got it. Thank you!

romandev (approved):
lgtm

PR #655, opened 2020-09-02
Fix a bug that the media in playing is paused when entering background

romandev:
The issue is because the media-suspend flag is renamed in upstream side.
  - https://chromium-review.googlesource.com/c/chromium/src/+/2144819

Refs: none


codeimpl (approved):
lgtm

jdragonbae (approved):
lgtm

PR #656, opened 2020-09-03
Fix a bug that the play state is not updated

romandev:
This is a regression after #651. Users can switch on/off fullscreen
mode, so play state should be synced regardless of fullscreen mode.

Refs: #567


codeimpl (approved):
lgtm

jdragonbae:
lgtm

PR #657, opened 2020-09-03
Fix a bug that the play() doesn't work on media remote if autoplay off

romandev:
It is because the play button on media remote is implemented using
MediaCommandProcesssor that internally calls `<video>.play()` in JS
side. So, this patch fixes the problem by making an exception case when
the MediaRemote is enabled and users already entered fullscreen mode.

Refs: #535


codeimpl (approved):
lgtm

PR #660, opened 2020-09-03
Make the pip button invisible on remote view

jdragonbae:
This patch means making the pip button invisible on the remote view.
Cause, it's only supported higher than O OS.

Refs: none


romandev (approved):
lgtm

PR #662, opened 2020-09-03
Implement RotationManager class to detect device rotation state

romandev:
This patch implements RotationManager class to detect device rotation
state including orientation state. Especially, if we use this class,
we can listen to the rotation/orientation status in non-activity class.

Refs: #658


jdragonbae (approved):
lgtm

PR #670, opened 2020-09-04
Make the content view re-layout whenever the device rotation is changed

romandev:
This patch makes the content view re-layout using RotationManager
whenever the device rotation/orientation is changed.

Although the rotation manager detects the rotation and orientation
information of the user device correctly, the activity configuration's
orientation might be updated late. It might cause breaking layout.
So, this patch postpones layouting the content view to the follow task.

Refs: #658


jdragonbae (approved):
lgtm

Issue #665, opened 2020-09-03
After saving naver credential with secure login, no username empty show on the suggestion


jdragonbae:
Tested on M OS (Galaxy S6)
4.01@85.0.4183.81 / debug / 64bit

**Steps to reproduce the problem:**
1. Go to https://nid.naver.com/nidlogin.login
2. Try to login
3. Save with secure login
4. Logout
5. Go to https://nid.naver.com/nidlogin.login
6. Click the id input field

**What is the expected behavior?**
Show suggestion with id

**What went wrong?**
Show suggestion with "no username"

romandev:
Uploaded upstream change: https://chromium-review.googlesource.com/c/chromium/src/+/2392191

romandev:
Fixed by #671 but keep to open this issue until upstream landed

PR #671, opened 2020-09-04
Fix a bug that the username is not stored on naver.com

romandev:
Save password prompt isn't shown on naver.com. The problem is
because the ID field value is reset when submitting the login form. At
this time, the value might be empty depending on renderer's IPC timing.
So, this patch makes it use typed_value if possible.

Refs: #665


jdragonbae (approved):
lgtm

PR #672, opened 2020-09-04
Fix a bug that the view state is not updated after re-layout

romandev:
This issue occurs because the re-layout operation initializes the view
states. This patch fixes the issue by caching the model data in the last
onUpdate() and then callling onUpdate() with the cached data again when
re-layout.

Refs: #658


romandev:
@jdragonbae PTAL?

jdragonbae (approved):
LGTM

romandev:
@jdragonbae Thanks!

Issue #675, opened 2020-09-04
The controls sometimes disappears too quickly


romandev:
**Reproducible steps**
1. Run Banana Browser
2. Play any `<video>` content
3. Enter fullscreen mode with `Media Remote` feature
4. Exit fullscreen mode
5. Re-enter fullscreen mode with `Media Remote` feature
6. Repeat 3-5 steps

**Expected**
When entering the fullscreen mode with `Media Remote` feature, the controls should be shown and it disappears after 5 seconds.

**Actual**
The controls disappears too quickly

PR #676, opened 2020-09-04
Fix a bug that the controls is sometimes hidden too quickly

romandev:
The issue is caused by `mWasControlsVisible` not being initialized. This
patch fixes the issue resetting the `mWasControlsVisible` when exit
fullscreen mode.

Refs: #675


jdragonbae (approved):
lgtm

Issue #677, opened 2020-09-04
There is an issue that the visibility of the controls are poor


romandev:
Currently, the controls background on media remote view is very nice, but there is a problem that the controls are not visible well in some environments(e.g. when video content's background color is white tone).

PR #678, opened 2020-09-04
Adjust the controls background color

romandev:
Currently, the controls background on media remote view is very nice,
but there is a problem that the controls are not visible well in some
environments(e.g. when video content's background color is white tone).
So, this patch removes the background gradation and applies a black
transparent dimming.

Refs: #677


jdragonbae (approved):
lgtm

PR #680, opened 2020-09-05
Fix a bug that the session doesn't keep for 60 seconds

romandev:
The session should keep for 60 seconds, but it is released on the third
authentication. It seems like a minor mistake. We should reset the
`mLastAuthenticationTime` if the session is expired.

Refs: #565


jdragonbae (approved):
lgtm

Issue #681, opened 2020-09-05
Enabling the browser lock feature triggeres ApplicationStatus.FOREGROUND


romandev:
**Reproducible steps**
1. Run Banana Browser
2. Visit Extension Settings
3. Enable Browser Lock feature
4. Click back button (to exit Extension Setting)
5. The `ApplicationStatus.FOREGROUND` event is triggered. It also triggers an authentication.

**Expected**
The `ApplicationStatus.FOREGROUND` should not be triggered.

**Actual**
The `ApplicationStatus.FOREGROUND` event is triggered. It also triggers an authentication.

knj0426:
#686

PR #682, opened 2020-09-05
Fix a bug that the unexpected ApplicationStatus.FOREGROUND is triggered

romandev:
The problem is because the `ApplicationStatusTracker` is started in
`BrowserLock.start()`. But the `ApplicationStatusTracker` should be
started when our actual browser application is launched, regardless of
`BrowserLock` module.

Refs: #681


knj0426:
LGTM, but #681 can resolved with #686 easily. How do you think?

romandev:
@knj0426, Thank you for review and your patch. In my humble opinion, your solution is simple but it doesn't work well on first run on memory as follows:

**Reproducible steps**
1. Enable Browser Lock feature
2. Exit Banana Browser (Remove it from recent app list)
3. Run Banana Browser

**Expected**
Authentication should be required

**Actual**
No authentication

knj0426 (approved):
LGTM

Issue #667, opened 2020-09-04
Volume, brightness seekbar not visible on N OS


jdragonbae:
Tested on N OS (Galaxy S6)
4.01@85.0.4183.81 / debug / 64bit

The icon(volume,  is visible. But, seekbar is not visible.

PR #683, opened 2020-09-05
Implement VerticalSeekBar and replace the rotated SeekBar

romandev:
The roated SeekBar has an issue that doesn't display properly on some
platforms(e.g. N-OS). So, this patch implements VerticalSeekBar and
replace the rotated SeekBar with new thing.

Refs: #667


jdragonbae (approved):
lgtm

PR #686, opened 2020-09-06
Set application status as FOREGROUND when enable BrowserLock

knj0426:
Initial status value is UNSPECIFIED so after enabling BrowserLock,
if user back to main activity, is triggers FOREGROUND event.
So, when enable BrowserLock, mCurrentStatus should be set to FOREGROUND.

Refs: #681

### DIFF
--- a/src/org/triple/banana/lock/ApplicationStatusTracker.java
+++ b/src/org/triple/banana/lock/ApplicationStatusTracker.java
@@ -68,6 +68,7 @@ public class ApplicationStatusTracker {
     }
 
     public void start() {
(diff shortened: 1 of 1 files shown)


romandev:
@knj0426, Thank you for review and your patch. In my humble opinion, your solution is simple but it doesn't work well on first run on memory as follows:

**Reproducible steps**
1. Enable Browser Lock feature
2. Exit Banana Browser (Remove it from recent app list)
3. Run Banana Browser

**Expected**
Authentication should be required

**Actual**
No authentication

Issue #687, opened 2020-09-06
BrowserLock might start() after authentication for SecureLogin


romandev:
**Reproducible steps**
1. Only register KeyGuard (without Biometrics) on system settings
2. Run Banana Browser
3. Store ID/PW on any site (e.g. https://rsolomakhin.github.io/autofill)
4. Visit Extension Settings
5. Enable Secure Login, Disable Browser Lock feature
6. Exit Banana Browser (remove it from recent app list)
7. Run Banana Browser again
8. Visit Extension Settings
9. Enable Browser Lock feature
10. Try to sign-in via Secure Login on the web site used in step 2.

**Expected**
The authentication should be required only once.

**Actual**
The authentication occurs twice (for SecureLogin and BrowserLock)

PR #688, opened 2020-09-06
Fix a bug that BrowserLock starts after authentication for SecureLogin

romandev:
The keyguard authentication runs as a separated system activity from our
application. So, while the keyguard authentication is in progress, our
application enters the background state. Then, after authenticaiton, the
foreground event is fired again, and then it triggers the browser lock.
So, this patch fixes the problem to pause the browser lock during the
keyguard authentication.

Refs: #687


jdragonbae (approved):
lgtm

romandev (commented):
This patch needs a more investigation.

Issue #482, opened 2020-07-09
Disable feed page in m84


romandev:
After rebase m84, the `enable_feed_in_chrome = false` causes build errors. This is a regression after rebase.

romandev:
@codeimpl any update?

romandev:
@codeimpl If you don't have enough time to investigate this issue, I'll take this.

PR #689, opened 2020-09-06
Disable feed suggestions in new tab page

romandev:
We've used `enable_feed_in_chrome = false` build flag to disable the
feed suggestion on new tab page. But this way is no longer available due
to upstream changes. So this patch disables the feature by modifying
preference value at runtime.

Refs: #482


codeimpl (approved):
LGTM

Issue #690, opened 2020-09-06
Cannot launch app if pattern lock is deleted when try to authenticate BrowserLock with pattern lock


knj0426:
1. BrowserLock with pattern lock
2. Sent app to background
3. Session expired
4. try to launch app
5. Authentication failed immediately because FallbackBackend return fail.
Before authentication BrowserLock should check device is secured or not

PR #693, opened 2020-09-08
Check whether device is secured or not befor try to authenticate

knj0426:
While app is at background, if device is lock deleted, Authenticator
uses  FallbackBackend.
If user try to authenticate, FallbackBackend returns fail immediately.
So, BrowserLock should check whether device is secured or not bebore
create authenticator.

Refs: #690


romandev (approved):
lgtm

Issue #692, opened 2020-09-08
The media control buttons on the media remote doesn't work on some websites


romandev:
It doesn't work at some sites (e.g. 나무티비)

PR #694, opened 2020-09-08
Fix a bug that the media remote buttons don't work on some websites

romandev:
We implemented MediaCommandProcessor in JavaScript for media control in
MediaRemote. But this only works for the fullscreenElement of the top
frame. So, this patch fixes the issue by reimplementing the
MediaCommandProcessor using MediaSession APIs.

Refs: #692


jdragonbae (approved):
lgtm!

PR #695, opened 2020-09-08
Fix two critical warnings

romandev:
This patch fixes the following critical warnings:

```
../../triple_banana/src/org/triple/banana/authentication/BiometricPromptActivity.java:59:
warning: [FallThrough] Execution may fall through from the previous case;
add a `// fall through` comment before this line if it was deliberate
                case ERROR_LOCKOUT_PERMANENT:
                ^
    (see https://errorprone.info/bugpattern/FallThrough)

../../triple_banana/src/org/triple/banana/base/widget/VerticalSeekBar.java:39:
warning: [UnsynchronizedOverridesSynchronized] Unsynchronized method
onDraw overrides synchronized method in AppCompatSeekBar
    protected void onDraw(Canvas c) {
                   ^
    (see https://errorprone.info/bugpattern/UnsynchronizedOverridesSynchronized)
```

Refs: none


knj0426 (approved):
lgtm

PR #696, opened 2020-09-08
Fix a bug that the media remote is sometimes not closed

romandev:
There is a bug that the media remote is sometimes not closed. So, this
patch makes the back button click dismiss the media remote view by
force.

Refs: #489


romandev:
TBR: @codeimpl, @jdragonbae, @knj0426

PR #699, opened 2020-09-08
Fix a crash when entering fullscreen without blink features flag

romandev:
Refs: #489


romandev:
TBR: @codeimpl, @jdragonbae, @knj0426

PR #706, opened 2020-09-09
Show toast when try to enter pip mode not in playing

jdragonbae:
This patch means showing the toast if a user try to enter pip mode while
in the video status is not playing.

Refs: #703


romandev (approved):
LGTM

PR #713, opened 2020-09-10
Add authentication description for browser lock

jdragonbae:
The authentication module called from the browser lock will use
`browser_lock_description`. Currently, it's determined by the value of
`isBackground` which only true by browser lock.

Refs: none


romandev (commented):
@jdragonbae Could you update this to biometric prompt backend?

src/org/triple/banana/authentication/BiometricPromptBackend.java
```diff
@@ -49,11 +49,14 @@ public void authenticate(boolean isBackground, @NonNull Callback callback) {
     }
 
     private void authenticateInternal(FragmentActivity parent) {
+        // FIXME(#711): Consider a new way to change the biometric prompt builder description
+        boolean isBackground = mBackground != null && mBackground.isShowing();
```
romandev, line 53: Should be
```
boolean isBackground = mBackground != null;
```
Because the mBackground might be destroyed for some reason?

jdragonbae: Thanks for the review. Done.

romandev (approved):
lgtm

Issue #714, opened 2020-09-10
A crash occurs sometimes when entering fullscreen mode with MediaRemote


romandev:
**Reproducible steps**
1. Run Banana Browser
2. Access youtube.com
3. Enter fullscreen mode
4. Exit Banana Browser using Recent Key
5. Repeat 1-4 steps

**Callstacks**
```
09-10 23:53:50.876  3558  3558 W System.err: android.view.WindowManager$BadTokenException: Unable to add window -- token android.os.BinderProxy@ed67452 is not valid; is your activity running?
09-10 23:53:50.876  3558  3558 W System.err:    at android.view.ViewRootImpl.setView(ViewRootImpl.java:1056)
09-10 23:53:50.876  3558  3558 W System.err:    at android.view.WindowManagerGlobal.addView(WindowManagerGlobal.java:381)
09-10 23:53:50.877  3558  3558 W System.err:    at android.view.WindowManagerImpl.addView(WindowManagerImpl.java:93)
09-10 23:53:50.877  3558  3558 W System.err:    at android.app.Dialog.show(Dialog.java:470)
09-10 23:53:50.877  3558  3558 W System.err:    at org.triple.banana.media_remote.MediaRemoteViewImpl.show(MediaRemoteViewImpl.java:219)
09-10 23:53:50.877  3558  3558 W System.err:    at org.triple.banana.media_remote.MediaRemoteService$1.onEnteredVideoFullscreen(MediaRemoteService.java:79)
09-10 23:53:50.877  3558  3558 W System.err:    at org.triple.banana.media.MediaEventDispatcher.onEnteredVideoFullscreen(MediaEventDispatcher.java:67)
09-10 23:53:50.877  3558  3558 W System.err:    at org.triple.banana.media.mojom.BananaMediaEventDispatcher_Internal$Stub.accept(BananaMediaEventDispatcher_Internal.java:222)
09-10 23:53:50.877  3558  3558 W System.err:    at org.chromium.mojo.bindings.RouterImpl.handleIncomingMessage(RouterImpl.java:238)
09-10 23:53:50.877  3558  3558 W System.err:    at org.chromium.mojo.bindings.RouterImpl.access$000(RouterImpl.java:21)
09-10 23:53:50.877  3558  3558 W System.err:    at org.chromium.mojo.bindings.RouterImpl$HandleIncomingMessageThunk.accept(RouterImpl.java:33)
[...]

PR #716, opened 2020-09-10
Fix a crash when entering fullscreen mode with MediaRemote

romandev:
A crash occurs sometimes when entering fullscreen mode with MediaRemote.
It is caused when the parent of the dialog is no longer valid. In the
current implementation, we are caching the dialog instance and then
re-using it in media remote view. But if the parent is no longer valid,
it causes a crash when the mDialog.show() is called. Moreover, the
parent activity's instance might be changed in some cases, so, we should
recreate it whenver the show() method in media remote view is called.

Refs: #714


jdragonbae (approved):
lgtm

codeimpl:
lgtm

Issue #712, opened 2020-09-10
Browser lock is broken sometimes


romandev:
**Reproducible steps**
1. Enable Browser Lock at Extension Settings
2. Run Banana Browser
3. Exit Banana Browser
4. Repeat 2-3 steps until reproduced

**Expected**
Browser lock should work well. Especially, if user's authentication failed, then should exit Banana Browser.

**Actual**
Browser lock is broken without authentication

PR #719, opened 2020-09-12
Refactor ApplicationStatusTracker

romandev:
The current implementation is using ApplicationStateListener and
ActivityStateListener to detect the application foreground/background
state. This patch removes ApplicationStateListener and timing hack, and
use hasVisibleActivities() instead.

Refs: #712


romandev:
TBR

PR #720, opened 2020-09-12
Introduce shutdown() API in BananaApplicationUtils

romandev:
Use shutdown() instead of moveTaskToBack(true);

Refs: #712


romandev:
TBR

PR #722, opened 2020-09-12
Implement BiometricPromptBackend and use it

romandev:
This patch implements the new BiometricPromptBackend. In case of the
similar existing BiometricPromptActivity, the mediator activity is
provided between the authentication requester and the authenticator.
However, there is a bug that the activity would be finished due to
unexpected activity life cycle. Especially, it is reproducible by
ChromeLauncherActivity and then it eventually causes BrowserLock broken.

Refs: #721


romandev:
TBR

romandev:
Reference link should be #712

PR #723, opened 2020-09-12
Introduce ApplicationStatusTracker.reset() and use it

romandev:
The ApplicationStatusTracker's event is triggered asynchronously.
Therefore, if the app restarts quickly, the background/foreground events
might not be detected. In this case, the BrowserLock might be broken
because the foreground event is not detected when re-entering the app
again. So, we should reset the state explicitly in this case.

Refs: #712


romandev:
TBR

Issue #725, opened 2020-09-17
GestureDetector operation error


codeimpl:
When the statusbar or navigationbar is operated, the gestures operate incorrectly.

PR #726, opened 2020-09-17
Fix getstureDetector operation error

codeimpl:
When statusbar or navigationbar is operated, the gestureDetector operate
incorrectly. This patch add prevention margin at touch region.

Refs: #725


codeimpl:
We need to discus whether it is best way or not.

res/layout/media_remote_view.xml
```diff
@@ -4,12 +4,20 @@
      License, v. 2.0. If a copy of the MPL was not distributed with this
      file, You can obtain one at https://mozilla.org/MPL/2.0/. -->
 
-<org.triple.banana.media_remote.MediaRemoteLayout 
+<RelativeLayout
```
romandev, line 7: Nested layout causes performance degradation.

romandev (commented):
I think you could early return onScroll() method calculating percentage

PR #728, opened 2020-09-19
Revert "Introduce shutdown() API in BananaApplicationUtils (#720)"

romandev:
This reverts commit e8fa4ce99ca0a217a32e907bdf23e5788188fa10.

Reason for revert: shutdown() API sometimes causes abnormal exit.

Original change's description:
> Introduce shutdown() API in BananaApplicationUtils (#720)
>
> Use shutdown() instead of moveTaskToBack(true);
>
> Refs: #712

Refs: #712


romandev:
TBR

PR #730, opened 2020-09-19
Implement KeyguardBackend and use it

romandev:
This patch implements the new KeyguardBackend. In case of the similar
existing KeyguardActivity, the mediator activity is provided between the
authentication requester and the authenticator. However, there is a bug
that the activity would be finished due to unexpected activity life
cycle. Especially, it is reproducible by ChromeLauncherActivity and then
it eventually causes BrowserLock broken.

Refs: #712


romandev:
TBR

PR #731, opened 2020-09-19
Remove legacy activity based backend implementation

romandev:
We've already switched to a new implementation. Therefore, this patch
removes the legacy implementation.

Refs: #712


romandev:
TBR

Issue #717, opened 2020-09-10
A crash occurs when a user click cancel button on browser lock authentication


jdragonbae:
**Reproducible steps**
1. Register fingerprint & pattern on M OS device
2. Install banana browser
3. Run banana browser
4. Go to the banana extension
5. Switch on browser lock
6. Click home button
7. Re-run banana browser
8. Click the cancel button on the authentication dialog

**Callstacks**
```
09-11 01:00:51.442 28446 28446 E AndroidRuntime: java.lang.AssertionError
09-11 01:00:51.442 28446 28446 E AndroidRuntime: 	at org.triple.banana.authentication.Authenticator.authenticateWithKeyguardAsFallback(Authenticator.java:41)
09-11 01:00:51.442 28446 28446 E AndroidRuntime: 	at org.triple.banana.authentication.BiometricPromptActivity$AuthenticationCallback.onAuthenticationError(BiometricPromptActivity.java:58)
09-11 01:00:51.442 28446 28446 E AndroidRuntime: 	at androidx.biometric.FingerprintHelperFragment$1$1.run(FingerprintHelperFragment.java:123)
09-11 01:00:51.442 28446 28446 E AndroidRuntime: 	at android.os.Handler.handleCallback(Handler.java:751)
09-11 01:00:51.442 28446 28446 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:95)
09-11 01:00:51.442 28446 28446 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:154)
09-11 01:00:51.442 28446 28446 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:6682)
09-11 01:00:51.442 28446 28446 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
09-11 01:00:51.442 28446 28446 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1520)
09-11 01:00:51.442 28446 28446 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1410)
```

**Expected**
A crash NEVER occurs.

**Actual**
A crash occurs.

jdragonbae:
It seems to be fixed by #722.

PR #732, opened 2020-09-20
Make the authenticate() method take parent activity

romandev:
Currently, the authenticate() method internally uses
getLastTrackedFocusedActivity() to find the parent activity for
convenience. This is no problem in the general case, but in the case of
BrowserLock feature, it might has a problem due to subtle timing issue.
So, this patch adds a new version of authenticate() that takes the
parent activity explicitly.

Refs: #712, #717


romandev:
TBR

PR #733, opened 2020-09-20
Reduce the touch bounds to prevent unexpected gesture detection

codeimpl:
Before this patch, the touch bounds is the same with the target view's
drawing bounds. So, it causes the media gestures to conflict with
opening status bar or navigation bar. To prevent the unexpected gesture
detection, this patch reduces the touch bounds. After this patch, the
bounds is reduced to 90%.

Refs: #725
Co-authored-by: Jinho Bang <zino@chromium.org>


romandev (changes requested):
Not LGTM; This is not working. Build failure.

src/org/triple/banana/media_remote/MediaRemoteGestureDetector.java
```diff
@@ -84,6 +84,17 @@ private GestureDetectorCompat createGestureDetectorCompat() {
...
+                        int noTouchHorizontalRegion = mTargetView.get().getWidth() * 0.05f;
+                        int noTouchVerticalRegion = mTargetView.get().getHeight() * 0.05f;
+
+                        if ((noTouchVerticalRegion > e.getY())
+                                || (noTouchHorizontalRegion > e.getX()
+                                           || (mTargetView.get().getWidth()
+                                                   - noTouchHorizontalRegion))
+                                        < e.getX()) {
+                            return false;
+                        }
```
jdragonbae, line 96: Is this mean the below conditions?

`
noTouchVerticalRegion > e.getY() || 
`
`
noTouchHorizontalRegion > e.getX() || 
`
`
mTargetView.get().getWidth() - noTouchHorizontalRegion < e.getX()
`

codeimpl: yes right, I am fixed it.

codeimpl:
> Not LGTM; This is not working. Build failure.

Sorry, I updated it again. plz review it.
thanks.

src/org/triple/banana/media_remote/MediaRemoteGestureDetector.java
```diff
@@ -84,6 +84,15 @@ private GestureDetectorCompat createGestureDetectorCompat() {
                 new GestureDetector.SimpleOnGestureListener() {
                     @Override
                     public boolean onDown(MotionEvent e) {
+                        int noTouchHorizontalRegion = (int) (mTargetView.get().getWidth() * 0.05f);
```
romandev, line 87: Should add a null check for `mTargetView` and `mTargetView.get()` because it is a weak reference.

src/org/triple/banana/media_remote/MediaRemoteGestureDetector.java
```diff
@@ -84,6 +84,15 @@ private GestureDetectorCompat createGestureDetectorCompat() {
...
+                        int noTouchVerticalRegion = (int) (mTargetView.get().getHeight() * 0.05f);
+
+                        if (noTouchVerticalRegion > e.getY() || noTouchHorizontalRegion > e.getX()
```
romandev, line 90: I'm not sure but shouldn't we check all bounds? This is missing bottom boundary check.

src/org/triple/banana/media_remote/MediaRemoteGestureDetector.java
```diff
@@ -84,6 +84,15 @@ private GestureDetectorCompat createGestureDetectorCompat() {
...
+                                || mTargetView.get().getWidth() - noTouchHorizontalRegion
+                                        < e.getX()) {
```
romandev, line 92: If we use `Rect` class, this is more readable and easier.

romandev:
I've just updated using `Rect`

romandev (approved):
lgtm

Issue #715, opened 2020-09-10
Unnecessary authentication request for BrowserLock when toggling dark mode


romandev:
**Reproducible steps**
1. Run Banana Browser
2. Visit Extension Settings
3. Disable Browser Lock
4. Enable Browser Lock
5. Enable Dark Mode

**OS**
Android P is reproducible

**Expected**
Authentication request is unnecessary.

**Actual**
Requires authentication for Browser Lock

PR #734, opened 2020-09-20
Introduce pauseForAMoment() method to BrowserLock

romandev:
This patch introduces pauseForAMoment(), which pauses the browser lock
feature temporarily for some reasons. This fixes the unexpected
authentication request issue when using Secure Login via KeyguardBackend
or toggling dark mode.

Refs: #715


romandev:
TBR

PR #735, opened 2020-09-20
Refactor media remote feature in engine side

romandev:
This patch refactors media remote feature code in engine side. This
especially fixes the following problems:
  - Fix a bug that the current time is randomly updated during position
    changed
  - Fix a bug that the play/pause doesn't work well in some case
  - Fix a bug that the media remote view appears duplicately in
    animeseries.io

Refs: none


romandev:
TBR

Issue #737, opened 2020-09-21
Media remote is not working for basic video


romandev:
**Reproducible steps**
1. Run Banana Browser
2. Visit https://www.w3schools.com/html/html5_video.asp
3. Make the video enter fullscreen mode

**Expected**
The video should be controlled by media remote feature

**Actual**
The media remote's view is shown but the video is not controlled

PR #738, opened 2020-09-21
Fix a bug that the media remote is not working for basic video

romandev:
The issue is caused that the querySelector() could only children element
except itself. If the pending element is video element itself, it
couldn't find it. So, this patch fixes the problem by adding IsA()
check.

Refs: #737


jdragonbae (approved):
lgtm

Issue #744, opened 2020-09-24
When a user click the home button, fullscreen mode is not exit fully


jdragonbae:
It happens on N OS device (not supported pip mode).

PR #746, opened 2020-09-24
Fix a bug when a user click home button on media remote

jdragonbae:
This patch means fixing the bug when a user click home button on media
remote. The problem is fullscreen mode is not exit fully. It happens
under the N OS(not supported pip mode).

Refs: #744, #745


romandev:
Fixed by #752

Issue #747, opened 2020-10-04
Sometimes app freeze occurs on Android 10


romandev:
When playing media, sometimes app freeze occurs with ARM64 apk on Android 10. The cause of this problem is unclear, but until fixing it, we should load the 32bit native library instead of 64bit.

PR #748, opened 2020-10-04
Enable use32bitAbi on ARM64 build

romandev:
When playing media, sometimes app freeze occurs with ARM64 apk on
Android 10. The cause of this problem is unclear, but until fixing it,
we should load the 32bit native library instead of 64bit.

Refs: #747


romandev:
TBR: @codeimpl, @jdragonbae

PR #751, opened 2020-10-04
Fix the update_strings command bug

romandev:
This is a regression after #748. The duplicated resources are generated
in android_clang_arm directory. We should exclude them.

Refs: none


romandev:
TBR: @codeimpl, @jdragonbae

PR #752, opened 2020-10-04
Fix a bug when a user click home button on media remote on N- OS

romandev:
There is a bug that the media remote doesn't shutdown when pressing the
home button on Android N- OS(doesn't support PIP mode). This patch fixes
the problem by additionally checking the flag of the video element
registered in the FullscreenMediaController when the document's
fullscreen element is lost.

Refs: #744, #745


romandev:
TBR: @codeimpl, @jdragonbae

Issue #761, opened 2020-10-10
The subresource_filter doesn't work after app update


romandev:
There is a problem with subresource_filter not working at all when updating from v3.02 to v4.06. FYI, it works well with a new installation of 3.02 or 4.06. It is only reproducible on update case.

PR #762, opened 2020-10-10
Fix a bug that the subresource_filter doesn't work after app update

romandev:
There is a problem with subresource_filter not working at all when
updating from v3.02 to v4.06. FYI, it works well with a new installation
of 3.02 or 4.06. It is only reproducible on update case.
But we didn't find the root cause. So, this patch fixes the problem by
updating the filter if it is not latest version.

Refs: #761


codeimpl (approved):
LGTM

Issue #763, opened 2020-10-10
The browser lock is triggered even if the option is not enabled


romandev:
**Reproducible steps**
1. Install Banana Browser newly (Should clear all data)
2. Run Banana Browser
3. Toggle dark mode on Banana Extension Settings
4. Enter home screen (sending app to background)
5. Return back to the app

**Expected behavior**
The browser lock should not be triggered if the option is not enabled

**Actual behavior**
The browser lock is triggered

PR #764, opened 2020-10-10
Fix a bug that the browser lock is triggered even if it is disabled

romandev:
The pauseForAMoment() is missing the option flag check. So, even if the
option is enabled, the browser lock might be triggered. This patch fixes
the problem by adding a flag check.

Refs: #763


codeimpl (approved):
LGTM

PR #780, opened 2020-10-19
Add BananaJavaScriptCallback interface

jdragonbae:
This patch adds the BananaJavaScriptCallback which extends JavaScriptCallback
interface. This enables getting the result about executing the JavaScript.

Refs: #776


romandev (approved):
lgtm

PR #781, opened 2020-10-19
Move function to check youtube domain to common class

jdragonbae:
This patch means moving the function to check the youtube domain to the common
utils class. Cause, this function is useful to use in other code.

Refs: #776


romandev (changes requested):
The method doesn't have a dependency with upstream. Therefore, there is no reason to move it to Banana layer.

jdragonbae:
> The method doesn't have a dependency with upstream. Therefore, there is no reason to move it to Banana layer.

Ack. Thanks.

src/org/triple/banana/util/UrlUtil.java
```diff
@@ -0,0 +1,36 @@
...
+import android.text.TextUtils;
+import android.util.Log;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+public class UrlUtil {
+    private static final String TAG = "UrlUtil";
+
+    public static boolean isYoutubeDomainUrl(String urlString) {
```
romandev, line 17: Make the string non-null

src/org/triple/banana/util/UrlUtil.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.triple.banana.util;
+
+import android.text.TextUtils;
+import android.util.Log;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+public class UrlUtil {
```
romandev, line 14: Make this class's name `YouTubeUtil` or `YouTubeHelper` or other name?

jdragonbae: `YouTubeUtil` looks good to me!

src/org/triple/banana/util/UrlUtil.java
jdragonbae: Done.

romandev (approved):
LGTM

PR #773, opened 2020-10-14
Add Tab Switcher button on the bottom toolbar

codeimpl:
Tab Switcher button exists on the top toolbar. But it is not easy to
handle it. So this patch adds the button to the bottom toolbar.

Refs: #772
Co-authored-by: Jaeyong Bae <jdragon.bae@gmail.com>

PR #783, opened 2020-10-20
Fix a bug the upstream toolbar appears for a while

jdragonbae:
This patch means fixing a bug the upstream toolbar appears for a while. It's
regression after #773. At this patch, we moved creating toolbar buttons in
`buttonInitializeWithNative()` to set listener `mTabSwitcherListener`. This
changes the timing for `viewGroup.removeAllViews()`, it turns out the upstream
toolbar appears for a while. So, we just do set `mTabSwitcherListener` on
`buttonInitializeWithNative()`.

Refs: #782


romandev (approved):
LGTM

PR #786, opened 2020-10-23
Implement YouTube quality options skeleton code

jdragonbae:
This patch implements skeleton code about quality options for YouTube. It will
control the quality using JavaScript.

Refs: #776


src/org/triple/banana/media_remote/MediaRemoteViewImpl.java
```diff
@@ -224,6 +233,7 @@ public void dismiss() {
         if (mDialog == null || !mDialog.isShowing()) return;
 
         resetContentViewInteractionListeners();
+        YouTubeUtil.dismissQualityOptionsDialog();
```
romandev, line 236: Why do we need this? As I know, if the parent is gone, does the dialog is also dismissed automatically?

jdragonbae: When enter pip mode, the dialog remain on the media. WDYT?

romandev: Ah okay, I missed it.

romandev (approved):
lgtm

Issue #488, opened 2020-07-13
Support for ADM / IDM


romandev:
Many users want that our browser supports the ADM / IDM features.

PR #787, opened 2020-10-23
Implement DownloadInterceptor

romandev:
This is an initial work for ADM/IDM support. When the download request
is started, then hooks the request and then DownloadInterceptor
can process the its own work.

Refs: #488


jdragonbae (approved):
lgtm

PR #788, opened 2020-10-23
Implement external download manager supports

romandev:
This patch implements external download manager supports. The option is
default disable and if it is enabled, then external download manager is
triggered such as ADM, IDM.

Refs: #488


romandev:
PTAL

jdragonbae (approved):
lgtm!

Issue #791, opened 2020-10-26
Replace CLEAR_BROWSING_DATA with CLEAR_DATA


codeimpl:
CLEAR_BROWSING_DATA is too long, so it can't be displayed on toolbar editor. So it needs to replace shorter one.

romandev:
@codeimpl any update?

PR #792, opened 2020-10-27
Replace CLEAR_BROWSING_DATA with CLEAR_DATA

codeimpl:
CLEAR_BROWSING_DATA is too long to read in the bottom toolbar. This
patch replace it with CLEAR_DATA.

Refs: #791


romandev (approved):
lgtm

Issue #790, opened 2020-10-26
When clicking terminate button, clear browsing data


codeimpl:
Clear browsing data when user clicks terminate button on bottom toolbar.

romandev:
@codeimpl any update?

PR #793, opened 2020-10-27
Adding "clear brwosing data" logic on terminate

codeimpl:
If the automatic deletion option is turned on, the data should be
deleted when user click terminate button.

Refs: #790


banana_cake/java/org/banana/cake/CakeToolbarManager.java
```diff
@@ -170,6 +171,7 @@ public void openClearBrowsingDataPreference() {
     public void terminate() {
         ChromeTabbedActivity activity = getChromeTabbedActivity();
         if (activity == null) return;
+        BananaClearBrowsingData.get().clearBrowsingData();
```
jdragonbae, line 174: Shouldn't we do this when feature is enabled?
`if (ExtensionFeatures.isEnabled(FeatureName.AUTO_CLEAR_BROWSING_DATA))`

codeimpl: Yes, you are right. Plz review it again. thanks.

codeimpl:
1) When user click, delete data then call terminate func.
2) Need to check synchronous routine instead of asynchronous one.

banana_cake/interfaces/java/org/banana/cake/interfaces/BananaClearBrowsingData.java
```diff
@@ -5,10 +5,12 @@
...
 
+import org.chromium.chrome.browser.browsing_data.BrowsingDataBridge.OnClearBrowsingDataListener;
+
 public interface BananaClearBrowsingData {
     static BananaClearBrowsingData get() {
         return BananaInterfaceProvider.get(BananaClearBrowsingData.class);
     }
 
-    void clearBrowsingData();
+    void clearBrowsingData(OnClearBrowsingDataListener onClearBrowsingDataListener);
```
romandev, line 15: You don't need to expose the Chromium type to upper layer.
You should make the BananaOnClearBrowsingDataListener.

romandev (approved):
lgtm

Issue #796, opened 2020-10-28
Adblock doesn't work sometimes after app update


romandev:
**Reproducible steps**
1. Install `3.02` version
2. Check whether the adblock is working or not
3. Update `4.0x` version
4. Check whether the adblock is working or not

**Expected behavior**
It should work well even after app update.

**Actual behavior**
It doesn't work

PR #797, opened 2020-10-28
Reimplement subresource filter using JNI instead of mojom IPC

romandev:
This patch fixes the issue that the adblock doesn't work after app
update. We still don't know the exact root cause, but it might be
because of introducing `monochrome` in 4.0x (upgraded from 3.02).
This patch fixes the issue trying to update the filter ruleset by force
in a case of app update. To do it, this patch reimplements the
subresource filter module using JNI instead of mojom IPC (because the
app update detection logic is written using Android Java).

Refs: #796


jdragonbae (approved):
lgtm! :100:

PR #798, opened 2020-10-29
Change the order of available quality options

jdragonbae:
This patch changes the order of available quality options. `auto` option would
be default option when play a new media on YouTube. So, it should be top among
the available quality options.

Refs: #776


src/org/triple/banana/util/YouTubeUtil.java
```diff
@@ -187,6 +187,12 @@ public static void dismissQualityOptionsDialog() {
         try {
             JSONObject json = new JSONObject(jsonString);
             JSONArray availableQualities = json.getJSONArray("availableQualities");
+            for (int i = availableQualities.length() - 1; i > 0; i--) {
+                if (availableQualities.getString(i).equals("auto")) {
+                    availableQualities.put(0, availableQualities.remove(i));
```
romandev, line 192: This is O(n) operation.

src/org/triple/banana/util/YouTubeUtil.java
```diff
@@ -187,6 +187,12 @@ public static void dismissQualityOptionsDialog() {
         try {
             JSONObject json = new JSONObject(jsonString);
             JSONArray availableQualities = json.getJSONArray("availableQualities");
+            for (int i = availableQualities.length() - 1; i > 0; i--) {
```
romandev, line 190: The `auto` element might be first element but the loop is anyway O(n).
Also, The `availableQualities.put(0, availableQualities.remove(i));` is O(n).
So, your changes are O(n^2).

jdragonbae: Ack. Thanks for the review!

src/org/triple/banana/util/YouTubeUtil.java
jdragonbae: Ack.

romandev (approved):
lgtm

PR #799, opened 2020-10-29
Fix a bug that the subresource filter's ruleset data is downgraded

romandev:
If the installed filter version is higher than the remote filter
version, we should NOT try to install the remote filter.

Refs: #796


jdragonbae (approved):
lgtm!

PR #800, opened 2020-10-30
Fix a bug that the new feature icon is not displayed sometimes

romandev:
This is a regression after #797. It is because the application
initialization timing has changed(runNowOrAfterFullBrowserStarted()).
So, this patch makes the `setNewFeatureIcon(true)` change the visibility
of the new feature icon.

Refs: none


jdragonbae (approved):
lgtm

PR #801, opened 2020-10-30
Move the built-in filter version info to BUILD.gn

romandev:
The current implementation has the built-in filter version information
as constants in Java code. Therefore, we should have updated the version
information whenever the built-in filter is updated. This patch removes
the constants in Java code and then makes it generates the information
via GN.

Refs: #796


romandev:
TBR: @codeimpl, @jdragonbae

PR #804, opened 2020-10-30
Implement LongClickableSwitchPreference and use it for adblock pref

romandev:
We need a way to update the adblock filter ruleset by force. So, this
patch adds LongClickableSwitchPreference and use it adblock preference
in ExtensionFeatures.

Refs: #796


romandev:
TBR: @codeimpl, @jdragonbae

romandev:
Refs: #803

Issue #807, opened 2020-11-02
Build error occurs in clean build due to dependency cycle


romandev:
```
In file included from ../../components/download/internal/common/download_item_impl.cc:66:
In file included from ../../components/download/internal/common/android/download_collection_bridge.h:16:
In file included from ../../components/download/public/common/in_progress_download_manager.h:31:
In file included from ../../triple_banana/modules/public/interfaces.h:13:
In file included from ../../third_party/blink/public/platform/platform.h:51:
In file included from ../../third_party/blink/public/common/feature_policy/feature_policy.h:14:
../../third_party/blink/public/common/feature_policy/feature_policy_features.h:10:10: fatal error: 'third_party/blink/public/mojom/feature_policy/feature_policy_feature.mojom-forward.h' file not found
#include "third_party/blink/public/mojom/feature_policy/feature_policy_feature.mojom-forward.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

Issue #806, opened 2020-11-02
Build error occurs because BananaSwitchPreference.java is absent


romandev:
```
ninja: error: '../../triple_banana/banana_cake/interfaces/java/org/banana/cake/interfaces/BananaSwitchPreference.java', needed by 'obj/chrome/android/chrome_java.turbine.jar', missing and no known rule to make it
```

PR #808, opened 2020-11-02
Fix a build break in clear build

romandev:
The build errors occur for the following reasons:
  - Wrong dependency; The `interfaces.h` is used in all layers but it
    causes cycle dependencies. So, this patch removes the Blink related
    code from `interfaces.h` to resolve the cycle dependency. However,
    it is not fundamental solution and we should refactor the
    `interfaces.h` to have the correct dependencies.
  - `BananaSwitchPreference.java` is absent. This patch adds the file.

Refs: #806, #807


jdragonbae (approved):
lgtm

codeimpl (commented):
LGTM

Issue #809, opened 2020-11-02
Create preferences for new banana extension features


jdragonbae:
Create preferences for each according to new banana extension layout setting.

PR #811, opened 2020-11-03
Factor out custom preferences from ExtensionFeatures

romandev:
This patch factors out the custom preferences from ExtensionFeatures to
make them reusable in new settings implementation.

Refs: #809


jdragonbae (approved):
lgtm

romandev:
Duplicated with #818

PR #812, opened 2020-11-03
Fix a crash when running the new banana extension features

romandev:
We need to add settings.* to proguard exception list.

Refs: #809


jdragonbae (approved):
LGTM!

PR #805, opened 2020-11-02
Implement new extension preferences

codeimpl:
The Banana Extension list is getting longer and needs reorganization.
This patch reorganized the list by dividing it into detailed settings
and quick settings.

Refs: #809
Co-authored-by: Jaeyong Bae <jdragon.bae@gmail.com>

PR #814, opened 2020-11-04
Fix a bug that the adblock version is not displayed

romandev:
This is a regression since #805. The string ID has been changed.

Refs: #803, #809


jdragonbae (approved):
lgtm

PR #815, opened 2020-11-04
Disable NewExtensionFeatures settings

romandev:
Introduced NewExtensionFeatures but it is still unstable. So, this patch
disables the setting screen until it is getting stable by default.

Refs: #809


jdragonbae (approved):
lgtm

PR #816, opened 2020-11-04
Implement dialog util

jdragonbae:
This patch implements dialog util. It will have functions like
`showRestartDialog`.

Refs: #809


src/org/triple/banana/util/DialogUtil.java
```diff
@@ -0,0 +1,26 @@
...
+
+import android.content.Context;
+
+import androidx.appcompat.app.AlertDialog;
+import androidx.annotation.NonNull;
+
+import org.banana.cake.interfaces.BananaApplicationUtils;
+import org.triple.banana.R;
+
+public class DialogUtil {
```
romandev, line 16: If this is not used in general, there is no reason to separate it into the new DialogUtil class. I'm not sure but you could make a RestartPreference.

jdragonbae: SGTM! Thanks for the review. I reuploaded on #818.

PR #817, opened 2020-11-04
Remove setFilterVersion() method in VersionInfo

romandev:
The current implementation sets the version in the FilterLoader when the
ruleset filter is loadable. However, even if the ruleset installation
fails in native side, the version might be displayed as if it were
normal. So, this patch removes the setFilterVersion() and makes it call
the getVersion() in FilterLoader when it needs to display the version.

Refs: #803


src/org/triple/banana/version/VersionInfo.java
```diff
@@ -31,15 +32,11 @@ public String getVersionName() {
         return null;
     }
 
-    public static void setFilterVersion(String version) {
+    public static String getFilterVersion() {
+        String version  = FilterLoader.instance.getVersion();
```
jdragonbae, line 36: Need to be formatting.

src/org/triple/banana/version/VersionInfo.java
```diff
@@ -9,13 +9,14 @@
 import android.text.TextUtils;
 import android.util.Log;
 
+import org.triple.banana.adblock.FilterLoader;
```
jdragonbae, line 12: Need to be formatting.

src/org/triple/banana/version/VersionInfo.java
jdragonbae, line 36: ditto.

jdragonbae (approved):
LGTM. It seems to be worked!

PR #819, opened 2020-11-04
Introduce successCallback parameter to Banana/CakeSubresourceFilter

romandev:
We need a way to check if the ruleset data installation was successful.
This patch introduces the successCallback parameter to
Banana/CakeSubresourceFilter. FYI, note that the callback is only called
if the installation is successful. If the installation fails, then the
callback is never called.

Refs: #803


romandev:
TBR: @codeimpl, @jdragonbae

PR #820, opened 2020-11-04
Refactor FilterLoader to support the force ruleset update

romandev:
This patch refactors the FilterLoader class to support the force ruleset
update. This patch includes the following things:
  - s/filter/ruleset/gI
  - Expose three public methods:
    - forceUpdateRuleset(): This is called if the application is updated
      or users request to reset/update the ruleset by force(it can be
      triggered via long-click event on update ruleset preference).
    - forceUpdateRemoteRuleset(): This is called if users request to
      update the remote ruleset by force(it can be triggered via click
      event on update ruleset preference).
    - updateRulesetIfNeeded(): This is called if the application is
      launched.

Refs: #803


jdragonbae (approved):
lgtm

PR #823, opened 2020-11-05
Enable NewExtensionFeatures settings by default

romandev:
This patch enables the NewExtensionFeatures settings.

Refs: #809


jdragonbae (approved):
lgtm

PR #825, opened 2020-11-05
Fix a bug that the quick settings aren't synced with the detail settings

romandev:
This patch makes the ExtensionFeatureSettings refresh when onResume() is
triggered.

Refs: #809


codeimpl (approved):
LGTM

PR #827, opened 2020-11-05
Fix a crash that the "Privacy" menu on Settings

romandev:
The crash occurs due to the resource name conflicts. The "Privacy" menu
on Settings and "Privacy" menu on Banana Extension Settings are using
the same name. So, the resource is overriden and it causes a crash.

Refs: #809


jdragonbae (approved):
lgtm

PR #833, opened 2020-11-06
Reorganize the quick settings on Banana Extension

romandev:
There are too many items in the quick settings. So, this patch reduces
the items to improve usability.

Refs: #809


codeimpl (approved):
LGTM

PR #839, opened 2020-11-06
Implement subresource filter update UI

romandev:
This patch implements the subresource filter update UI.

Refs: #803


src/org/triple/banana/settings/preference/AdblockPreference.java
```diff
@@ -7,6 +7,7 @@
 
 import android.content.Context;
 import android.util.AttributeSet;
+import android.util.Log;
```
codeimpl, line 10: delete it :)

codeimpl (approved):
LGTM with nit

Issue #840, opened 2020-11-07
The android lint is not working


romandev:
Currently, the android lint is not working in [current code base](https://github.com/triplebanana/triple_banana/commit/299572bb6c93621967392d04621615e94bdde733)

PR #841, opened 2020-11-07
Fix a bug that the android lint doesn't work

romandev:
It might be due to upstream change. So, this patch cherry pick the
upstream CL[1] and manually set the enable_lint=true if it is debug
build.

[1] https://chromium-review.googlesource.com/c/chromium/src/+/2391077

Refs: #840


codeimpl (approved):
LGTM

PR #845, opened 2020-11-08
Extends RestartPreference for Media Features

jdragonbae:
The Media Features needs to be restart when it's changed. This patch
extends RestartPreference and makes this possible.

Refs: #843


romandev (approved):
LGTM. Thanks.

Issue #846, opened 2020-11-08
A crash occurs in 32bit apk


romandev:
```
11-08 22:23:28.880  8913  8913 W System.err: java.lang.UnsatisfiedLinkError: No implementation found for java.lang.String org.chromium.base.natives.GEN_JNI.org_banana_cake_CakeSubresourceFilter_getVersion() (tried Java_org_chromium_base_natives_GEN_1JNI_org_1banana_1cake_1CakeSubresourceFilter_1getVersion and Java_org_chromium_base_natives_GEN_1JNI_org_1banana_1cake_1CakeSubresourceFilter_1getVersion__)
11-08 22:23:28.881  8913  8913 W System.err:    at org.chromium.base.natives.GEN_JNI.org_banana_cake_CakeSubresourceFilter_getVersion(Native Method)
11-08 22:23:28.881  8913  8913 W System.err:    at org.banana.cake.CakeSubresourceFilterJni.getVersion(CakeSubresourceFilterJni.java:39)
11-08 22:23:28.881  8913  8913 W System.err:    at org.banana.cake.CakeSubresourceFilter.getVersion(CakeSubresourceFilter.java:27)
11-08 22:23:28.881  8913  8913 W System.err:    at org.triple.banana.subresource_filter.RulesetLoader.getVersion(RulesetLoader.java:220)
11-08 22:23:28.881  8913  8913 W System.err:    at org.triple.banana.subresource_filter.RulesetLoader.updateBuiltInRuleset(RulesetLoader.java:93)
11-08 22:23:28.881  8913  8913 W System.err:    at org.triple.banana.subresource_filter.RulesetLoader.updateRulesetIfNeeded(RulesetLoader.java:248)
11-08 22:23:28.881  8913  8913 W System.err:    at org.triple.banana.TripleBananaApplication.onInitialized(TripleBananaApplication.java:78)
11-08 22:23:28.881  8913  8913 W System.err:    at org.banana.cake.bootstrap.BananaApplication.lambda$attachBaseContext$0$BananaApplication(BananaApplication.java:24)
11-08 22:23:28.881  8913  8913 W System.err:    at org.banana.cake.bootstrap.BananaApplication$$Lambda$0.run(Unknown Source:2)
11-08 22:23:28.881  8913  8913 W System.err:    at org.chromium.chrome.browser.init.ChromeBrowserInitializer.onFinishFullBrowserInitialization(ChromeBrowserInitializer.java:397)
[...]

PR #847, opened 2020-11-09
Fix a crash when launching application with 32bit apk

romandev:
The crash occurs because the JNI native code in banana_cake/cpp is not
loaded in case of 32bit apk. We didn't find a root cause yet. So, this
patch fixes the problem by moving the cpp code to chrome's BUILD.gn as a
workaround.

Refs: #846


jdragonbae (approved):
lgtm

PR #848, opened 2020-11-09
Fix a bug that the BrowserLock is not working when app restarts

romandev:
The problem is because the ApplicationStatusTracker and BrowserLock
modules are initialized lazyily. The modules should be initialized
before onInitialized() is called. So, this patch adds
onBeforeInitialized() and moves the module initializer logic to the
method.

Refs: none


jdragonbae (approved):
lgtm

PR #849, opened 2020-11-09
Fix a bug that the dark mode switch is not synced

romandev:
Currently, there is a bug that if users change the dark mode via dark
mode button at bottom toolbar, the change is not synced with the dark
mode setting value on banana extension. The DarkModePreference sets the
initial value in constructor but the it can be overwritten in
onInitialValue(). So, this patch moves the initialization logic to
onInitialValue().

Refs: none


jdragonbae (approved):
lgtm

Issue #850, opened 2020-11-09
Touch to fill doesn't work after changing dark mode


romandev:
**Reproducible steps**
1. Visit https://rsolomakhin.github.io/autofill/
2. Store a credential at "Username/Password" section
3. Change dark mode (Settings - Theme - Dark Mode)
4. Try to autofill via "Touch to fill"
5. "Touch to fill" UI doesn't appear

**Expected behavior**
The "Touch to fill" UI should be shown

**Actual behavior**
The "Touch to fill" UI is not shown

romandev:
https://bugs.chromium.org/p/chromium/issues/detail?id=1146969

romandev:
https://chromium-review.googlesource.com/c/chromium/src/+/2525804

PR #851, opened 2020-11-09
Fix a bug that "Touch to fill" UI doesn't work when switching dark mode

romandev:
When switching dark mode, the application context is changed. Therefore,
existing cached object might not work. So, this patch removes the cache
related code and recreate the bridge object whenever
TouchToFillViewImpl::Show() is called.

Refs: #850


jdragonbae (approved):
lgtm

PR #856, opened 2020-11-09
Disable chromium crazy linker for 32bit binary

romandev:
This patch disables the chromium crazy linker for backward
compatibility.

Refs: #846


romandev:
TBR: @codeimpl, @jdragonbae

PR #859, opened 2020-11-09
Introduce disable_library_rename flag and use it

romandev:
The disable_library_rename flag was removed from upstream but it is
still needed in downstream. So, this patch restores the build flag and
use it instead of `use_chromium_linker=false`. This patch fixes a crash
with 32bit apk in L-OS.

Refs: #846


romandev:
TBR: @codeimpl, @jdragonbae

Issue #867, opened 2020-11-19
A crash occurs when changing "bottom toolbar visibility" on settings


romandev:
**Reproducible steps**
1. Clear all data
2. Run Banana Browser
3. Visit chrome://flags
4. Enable "Tab Groups"
5. Restart Banana Browser twice

**Expected behavior**
It should work well

**Actual behavior**
 A crash occurs as follows

```
11-19 10:46:49.722  3410  3410 D AndroidRuntime: Shutting down VM
11-19 10:46:49.722  3410  3410 E AndroidRuntime: FATAL EXCEPTION: main
11-19 10:46:49.722  3410  3410 E AndroidRuntime: Process: org.triple.banana, PID: 3410
11-19 10:46:49.722  3410  3410 E AndroidRuntime: java.lang.RuntimeException: Unable to start activity ComponentInfo{org.triple.banana/org.chromium.chrome.browser.ChromeTabbedActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'android.view.View android.view.ViewStub.inflate()' on a null object reference
11-19 10:46:49.722  3410  3410 E AndroidRuntime: 	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3113)
11-19 10:46:49.722  3410  3410 E AndroidRuntime: 	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3256)
11-19 10:46:49.722  3410  3410 E AndroidRuntime: 	at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:78)
11-19 10:46:49.722  3410  3410 E AndroidRuntime: 	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:108)
11-19 10:46:49.722  3410  3410 E AndroidRuntime: 	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:68)
11-19 10:46:49.722  3410  3410 E AndroidRuntime: 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1947)
11-19 10:46:49.722  3410  3410 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:106)
11-19 10:46:49.722  3410  3410 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:214)
[...]

PR #868, opened 2020-11-19
Fix a crash when enabling "Tab Groups" feature

romandev:
The crash occurs because the enableBottomToolbar() causes inflating
layout twice. So, this patch introduces "early return" if the
enableBottomToolbar() is already called.

Refs: #867


jdragonbae (approved):
lgtm

Issue #874, opened 2020-11-20
Resolve conflict with upstream "Secure DNS"


jdragonbae:
Currently, our implementation of "Secure DNS" has a conflict with the upstream side. 
After resolving this one, re-enable "Secure DNS" preference on Privacy Settings[1].

[1] chrome/android/java/src/org/chromium/chrome/browser/privacy/settings/PrivacySettings.java

Issue #872, opened 2020-11-20
Should disable "Secure DNS" menu on Privacy Settings


romandev:
The Secure DNS menu is introduced in upstream side but it has a conflict with our implementation. So, until fixing the problem, we should disable the menu on Privacy Settings.

PR #875, opened 2020-11-20
Disable "Secure DNS" menu on Privacy Settings

jdragonbae:
The Secure DNS menu is introduced in upstream side but it has a conflict
with our implementation. So, until fixing the problem, we should disable
the menu on Privacy Settings.

Refs: #872, #874


romandev (approved):
LGTM

Issue #873, opened 2020-11-20
The "New icon" is not shown sometimes after installation


romandev:
**Reproducible steps**
1. Install APK (5.x & 6.x are different)
2. Run "Banana Browser"
3. Check the "New Icon"
4. Repeat 1-3 steps

**Expected behavior**
The "New icon" is should be shown.

**Actual behavior**
The "New icon" is not shown sometimes after installation.

PR #876, opened 2020-11-25
Change check logic whether it is first install

jdragonbae:
This patch change the logic whether it is first install. We check it
using FirstRunFlowComplete which returns whether the main First Run
Experience flow is complete.

Refs: #873
Co-authored-by: Byoungkwon Ko <codeimpl@gmail.com>
Co-authored-by: Jinho Bang <zino@chromium.org>


codeimpl (approved):
lgtm

Issue #885, opened 2020-11-30
[Video Downloader] Implement Video Download UI on Media Remote


codeimpl:
We should implement Video Download UI on Media Remote. To resolve this issue, we should do the following things:
  - Implement UI Icon on Media Remote
  - Send state whether the playing video is downloadable from engine side to browser side

romandev:
Takes 02:20

romandev:
Takes **00:30** to make the video download button show only if the video is downloadable

romandev:
Takes 00:30 to implement the video download API

romandev:
Takes **03:20** totally.

codeimpl:
Thanks for your time update :+1:

PR #890, opened 2020-11-30
Implement video download UI on media remote

romandev:
This patch implements the video download UI on media remote.
  - Add vector icon
  - Provide model API to update whether to display icon or not.

Refs: #885


jdragonbae (approved):
lgtm

PR #891, opened 2020-11-30
Make the video download button show only if the video is downloadable

romandev:
We should make the video download button show only if the video is
actually downloadable. The downloadable state is coming from Blink and
passed to onEnteredVideoFullscreen() event.

Refs: #885


jdragonbae (approved):
lgtm

PR #893, opened 2020-11-30
Implement video download API for media command processor

romandev:
This patch implements download() API to download the playing video on
media remote. It is only works if the playing video is downloadable.

Refs: #885


jdragonbae (approved):
LGTM!

Issue #886, opened 2020-11-30
[QR Code] Implement QR code feature


codeimpl:
To resolve this issue, we should finish the following things:
- Add a new icon for QR code feature on bottom toolbar (1 hour)
- When clicks the QR code icon on bottom toolbar, show QRCode dialog using QRCodeScanCoordinator (1 hour)
  (NOTE: Make the QR code scanner show preferentially rather than QR code generator)

codeimpl:
Takes 0:30 for adding bottom toolbar, and takes 1:00 for qrCode dialog.

PR #895, opened 2020-11-30
Add QRCode feature for bottom toolbar

codeimpl:
This patch add QRCode feature to bottom toolbar. But this feature
location is just temporary, it can be changed soon.

Refs: #886


banana_cake/java/org/banana/cake/CakeToolbarManager.java
```diff
@@ -179,4 +180,13 @@ public void closeCurrentTab() {
         if (activity == null) return;
         activity.onMenuOrKeyboardAction(R.id.close_tab, false);
     }
+
+    @Override
+    public void openQRCodeDialog() {
+        ChromeTabbedActivity activity = getChromeTabbedActivity();
+        if (activity == null) return;
+        QrCodeCoordinator qrCodeCoordinator =
+                new QrCodeCoordinator(activity, activity.getActivityTab().getUrl().getSpec());
```
romandev, line 189: The getActivityTab() is nullable. Therefore, we should check whether it is null or not.

romandev: If the url is empty, what happens?

codeimpl: > The getActivityTab() is nullable. Therefore, we should check whether it is null or not.

done!

codeimpl: > If the url is empty, what happens?

qrcode will be null, it means that user can't get qrcode.

romandev (approved):
lgtm

Issue #889, opened 2020-11-30
[MediaRemote] Implement to control the playback rate


jdragonbae:
We will support playback speed in MediaRemote. For this one, we can utilize video playbackRate property in HTML DOM.
It should be implemented with UI to choose from.

jdragonbae:
Takes **01:00** for skeleton dialog code (#902).

jdragonbae:
Takes 02:00 for implement dialog code (#902).

PR #898, opened 2020-12-01
Implement to control playback rate on media remote

jdragonbae:
This patch implements playback rate to control playback rate on media
remote.

Refs: #889


jdragonbae:
@romandev , @codeimpl This is draft PR. Please, take a look if it is in the right way.

modules/public/mojom/media.mojom
```diff
@@ -18,11 +18,13 @@ interface BananaMediaEventDispatcher {
   OnChangedPIPMode(bool value);
   OnEnteredVideoFullscreen(pending_remote<BananaMediaCommandProcessor> impl, bool isDownloadable);
   OnExitedVideoFullscreen();
+  OnPlaybackRateChanged(double rate);
```
jdragonbae, line 21: Q. Do really need OnPlaybackRateChanged? Any other ideas?

romandev, line 21: Could you please explain why we should implement this event?

romandev: I thought that the event might be used for UI update, right?
I imagine that the we provide "playback rate list" on UI as follows:
  - 4x
  - 2x
  - 1x
  - 1/2x
  - 1/4x
If the video playback rate was changed to 0.01 (1 / 10), how can we update the UI?
(Other option might be that we implement slider UI)

jdragonbae: > I thought that the event might be used for UI update, right?

Right.

> I imagine that the we provide "playback rate list" on UI as follows:
> 
> 4x
> 2x
> 1x
> 1/2x
> 1/4x

It is the same as I thought. Similar to youtube quality.

> If the video playback rate was changed to 0.01 (1 / 10), how can we update the UI?
> Other option might be that we implement slider UI)

My idea now is to adjust the value. But, I'm not sure this way would be good.
Also, slider UI seems to be fine.

codeimpl: We think get() api is better than event handler.

romandev (commented):
As I remember, we decided to have a getter for playback rate, didn't it?

romandev (approved):
LGTM

PR #899, opened 2020-12-01
Introduce isPipSupported() in MediaRemoteViewModel and use it

romandev:
Remove the control logic from MediaRemoteViewImpl and move the logic to
MediaRemoteService.

Refs: none


jdragonbae (approved):
lgtm

Issue #888, opened 2020-11-30
[Secure DNS] Resolve conflict with our current implementation


jdragonbae:
Separately present in upstream and Banana Extension for Secure DNS preferences. We have to integrate this.

Through our analysis (#915), we decided behaviors as follows.

| State | Secure Mode | Secure Template | Preference value (secure_dns_switch) |
|-|-|-|-|
| Before first turn on Secure DNS | AUTO | "" | OFF |
| Turns on from off | ON | "1.1.1.1" | ON |

Except for the above two situations, it works the same as upstream Secure DNS.
Additionally, when turn on preference (secure_dns_switch), bypass is also turn on.

Case for app update situations, handling as below.

| feature_name_secure_dns | Secure Mode | Secure Template | Preference value (secure_dns_switch) |
|-|-|-|-|
| OFF | AUTO | "" | OFF |
| ON | ON | "1.1.1.1" | ON |

- [x] Actually in auto mode, If the switch has never been turned on, the preference is off
- [x] When the preference (secure_dns_switch) turns on from off, set mode (secure) and template ("1.1.1.1")
- [ ] Sync Banana Secure DNS switch preference on extension settings & the preference (secure_dns_switch)
- [ ] Implement preference UI in Privacy and Security
- [ ] Migration switch preference for app updates

jdragonbae:
Takes **01:00** for upstream changes.

jdragonbae:
Takes **02:00** for upstream changes (#903).

jdragonbae:
Takes **00:30** for migrating preferences (#905).

codeimpl:
Split this remaining works into #923, #924 and #925

PR #900, opened 2020-12-01
Implement banana cake local state 

jdragonbae:
This patch implements banana cake local state to get/set value from the
local state. It accesses the local state using the JNI Bridge.

Refs: #888

### DIFF
--- a/banana_cake/java/org/banana/cake/CakeLocalState.java
+++ b/banana_cake/java/org/banana/cake/CakeLocalState.java
@@ -1,0 +1,26 @@
+// Copyright 2020 The Triple Banana Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package org.banana.cake;
+
+import androidx.annotation.NonNull;
+
+import org.banana.cake.interfaces.BananaLocalState;
+
+import org.chromium.base.annotations.NativeMethods;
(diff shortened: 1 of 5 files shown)


jdragonbae:
@romandev, @codeimpl This is draft PR. I'm not sure whether I'm right. PTAL.

banana_cake/java/org/banana/cake/CakeLocalState.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.banana.cake.interfaces.BananaLocalState;
+
+import org.chromium.base.annotations.NativeMethods;
+import org.chromium.chrome.browser.preferences.Pref;
+
+public class CakeLocalState implements BananaLocalState {
+    @Override
+    public boolean isSecureDNSMode() {
+        return !CakeLocalStateJni.get().getString(Pref.DNS_OVER_HTTPS_MODE).equals("off");
```
romandev, line 18: I think you could just use `SecureDnsBridge.getMode();`, couldn't it?

jdragonbae: @romandev Thanks for the review. 

> I think you could just use SecureDnsBridge.getMode();, couldn't it?

Yes, we could use `SecureDnsBridge.getMode()` with access modifier change about `SecureDnsBridge`.
I'll try to change and upload it.

jdragonbae: I uploaded the new patch[1].

[1] https://github.com/triplebanana/triple_banana/pull/901

PR #903, opened 2020-12-03
Change the behavior related to upsteram Secure DNS

jdragonbae:
This patch does the following things:
* Set to Secure DNS off mode until the value has not been set by user.
It means that off mode in the UI, but it actually works as auto mode.

* Set Cloud Flare (1.1.1.1) as the service provider when turn on Secure
DNS from off.

* Remove the code related to hard coding (the option is available with
1.1.1.1).

Refs: #874, #888


upstream/tb_chrome_common_chrome_features.cc.patch
```diff
@@ -0,0 +1,35 @@
...
+@@ -295,8 +295,15 @@ const base::FeatureParam<std::string> kDnsOverHttpsDisabledProvidersParam{
+ 
+ // Set whether fallback to insecure DNS is allowed by default. This setting may
+ // be overridden for individual transactions.
+-const base::FeatureParam<bool> kDnsOverHttpsFallbackParam{&kDnsOverHttps,
+-                                                          "Fallback", true};
++const base::FeatureParam<bool> kDnsOverHttpsFallbackParam{
++  &kDnsOverHttps,
++#if defined(ENABLE_TRIPLE_BANANA)
++      "Fallback", false
```
romandev, line 13: Do we really need this? This makes always `secure` instead of `automatic`. If users want to use `automatic`, how can we support it?

jdragonbae: @romandev Thanks for the review. I was mistaken.
As you said, it always use Cloudflare service provider even though set `automatic`.
I changed this code.

upstream/tb_chrome_browser_net_stub_resolver_config_reader.cc.patch
```diff
@@ -0,0 +1,29 @@
...
+--- tb/chrome/browser/net/stub_resolver_config_reader.cc
++++ tb_diff/chrome/browser/net/stub_resolver_config_reader.cc
+@@ -136,6 +136,17 @@ StubResolverConfigReader::StubResolverConfigReader(PrefService* local_state,
+       }
+       default_doh_templates = features::kDnsOverHttpsTemplatesParam.Get();
+     }
++
++#if defined(ENABLE_TRIPLE_BANANA)
++    // Set the service provider as default when first running
++    if (local_state_->GetString(prefs::kDnsOverHttpsMode).empty()) {
```
romandev, line 11: Is this correct? This makes the SecureDNS enable by default.
We don't want to enable the feature by default.

romandev: Hmm.. I misunderstood this.
BTW, does the SecureDNS is enabled by default in upstream side?

jdragonbae: @romandev Yes, I think the feature[1] is enabled by default on the upstream side.

[1] https://source.chromium.org/chromium/chromium/src/+/master:chrome/common/chrome_features.cc;l=291

romandev (approved):
Okay, this is best way in current state. I looked into this part deeply but I have no better idea. Thanks. LGTM

PR #918, opened 2020-12-11
Implement the Secure DNS bridge to access config

jdragonbae:
This patch implements the Secure DNS bridge to get/set secure dns config
from the local state. It's used to synchronize the Secure DNS switch
preference values upstream and banana.

Refs: #888


codeimpl (approved):

upstream/tb_chrome_browser_privacy_java_src_org_chromium_chrome_browser_privacy_secure_dns_SecureDnsSettings.java.patch
```diff
@@ -10,6 +10,12 @@ diff --git tb/chrome/browser/privacy/java/src/org/chromium/chrome/browser/privac
 +                        "https://chrome.cloudflare-dns.com/dns-query" /* templates */,
 +                        true /* valid */));
 +            }
++            getContext()
++                    .getSharedPreferences("org.triple.banana_preferences", Context.MODE_PRIVATE)
```
romandev, line 14: This is not good. We should use `Banana` hook if needed.

romandev: I'm not sure but you could use the `isSecureDNSMode()` in preference change listener?

jdragonbae: @romandev Thanks for the review.  I was mistaken. I uploaded the patch that removed this code.

romandev (approved):
lgtm

Issue #883, opened 2020-11-30
[YouTube Caption] Implement the YouTube caption in engine side


codeimpl:
After analyzing/testing #882, we should implement YouTube caption supports in engine side actually. We might need refactoring the existing code.

codeimpl:
This feature is removed from v6.x because it has some side effects.

PR #922, opened 2020-12-13
Support closed caption for YouTube

romandev:
Refs: #883


romandev (commented):
This patch works anyway but we need more tests.

PR #942, opened 2020-12-16
Disable closed caption for YouTube feature by default

romandev:
This patch makes the CC feature for YouTube disable until it is
implemented completely.

Refs: #937


jdragonbae (approved):
lgtm

Issue #923, opened 2020-12-14
[Secure DNS] Sync Banana Secure DNS switch preference on extension settings & the preference (secure_dns_switch)


codeimpl:
.

PR #943, opened 2020-12-16
Set the Secure DNS mode when feature_secure_dns is on

jdragonbae:
This patch is for existing users using the Secure DNS mode. After the
app update, it maintains usability through upstream and Banana switch
preference sync.

Refs: #923


romandev (approved):
lgtm but you should check my comment in previous patch.

Issue #924, opened 2020-12-14
[Secure DNS] Implement preference UI in Privacy and Security


codeimpl:
.

PR #944, opened 2020-12-16
Add the Secure DNS preference on privacy and security

jdragonbae:
This patch means adding the Secure DNS preference. It's on the
privacy and security setting page and will open SecureDnsSettings
fragment when a user clicks this one.

Refs: #924


romandev (approved):
LGTM

PR #947, opened 2020-12-17
Use Map instead of List for ButtonInfo

jdragonbae:
This patch use Map instead of List for ButtonInfo. Cause, when a user
clicks QuickMenuButton, the QuickMenuController only knows ButtonId
clicked. Via this Id, the controller should be able to change the
ButtonInfo state in Map.

Refs: #933


romandev (commented):
Is this finished? I think you should implement the ID query how it works in this patch.

Issue #931, opened 2020-12-14
[QuickMenu] Implement container dialog


codeimpl:
We should implement the container dialog in `QuickMenuViewImpl`

PR #948, opened 2020-12-17
Implement QuickMenuDialog

codeimpl:
This patch implements QuickMenuDialog which is a container for QuickMenu.

Refs: #931


src/org/triple/banana/quick_menu/QuickMenuDialog.java
```diff
@@ -0,0 +1,41 @@
...
+import android.graphics.drawable.ColorDrawable;
+import android.view.Gravity;
+import android.view.Window;
+import android.view.WindowManager;
+
+import androidx.annotation.NonNull;
+
+import org.triple.banana.R;
+
+public class QuickMenuDialog extends Dialog {
```
jdragonbae, line 20: Can we change `public` to `package`?

codeimpl: done! :+1:

codeimpl:
@romandev @jdragonbae 
I added dialog reference at view.
Take a Look :)

res/layout/quick_menu_layout.xml
```diff
@@ -0,0 +1,19 @@
...
+     License, v. 2.0. If a copy of the MPL was not distributed with this
+     file, You can obtain one at https://mozilla.org/MPL/2.0/. -->
+
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:background="@drawable/quick_menu_background"
+    android:orientation="vertical">
+
+    <!-- It is just test view, it will be removed after review -->
```
romandev, line 13: Should we remove this for merging this patch?

codeimpl: @romandev 
do you mena the comment or view??

romandev: Your comment says that "it will be removed after review". If we merge this patch, should we remove this?

codeimpl: I think we can remove this because i am drawing view layout for the dialog.
I think it can be replaced the view with them.

romandev (approved):
lgtm

Issue #911, opened 2020-12-07
[Video Downloader] After downloading, popup appears behind MediaRemote.


codeimpl:
The popup that appear behind MediaRemote can't be touched or clicked by user. User should be able to handle the popup.

PR #950, opened 2020-12-21
Make the infobar hidden during fullscreen video

romandev:
This patch makes the infobar hidden during fullscreen video.

Refs: #911


jdragonbae (approved):
lgtm

PR #952, opened 2020-12-21
Enable the download button only when external download option is enabled

romandev:
This patch makes the download button enable only when the external
download manager is enabled. When downloading using the internal
download manager, there is a poor UX problem that the download progress
is not shown.

Refs: #911


jdragonbae (approved):
lgtm

Issue #897, opened 2020-12-01
[Toolbar] Toolbar editor layout is broken


romandev:
There are too many buttons on bottom toolbar editor. So, it is broken. This is a regression issue after #895.

romandev:
I've investigated this issue enough, but I could not be sharp. After discussion with our team, we decided to move the QR code button to QuickMenu, and then we can reduce the editable button's count.

PR #953, opened 2020-12-22
Adjust text size of toolbar editor description

romandev:
Because many features have been added in our browser, there are
currently too many candidate buttons on the bottom toolbar editor.
So, the layout is broken after #895 PR merged. So, this patch resolves
the issue by adjusting the text size of the toolbar editor description.

Refs: #897


codeimpl (approved):
LGTM

Issue #955, opened 2020-12-22
[QuickMenu] Refactor QuickMenu architecture


romandev:
References:
  - https://www.raywenderlich.com/34-design-patterns-by-tutorials-mvvm
  - https://www.objc.io/issues/13-architecture/mvvm/
  - https://proandroiddev.com/dialogs-in-android-mvvm-5d6e1ca53b19
  - http://www.plainionist.net/Mvvm-Dialogs/

PR #956, opened 2020-12-22
Factor out QuickMenuViewModelData from QuickMenuViewModel

romandev:
The nested class is a little complicated. So, this patch factors out
QuickMenuViewModelData from QuickMenuViewModel.

Refs: #955


jdragonbae:
The patch looks good to me. But I couldn't find the `QuickMenuViewModelData.java` file. 
Is there something I am missing?

romandev:
> The patch looks good to me. But I couldn't find the `QuickMenuViewModelData.java` file.
> Is there something I am missing?

You're right. I missed the file. Please check my new patch.

jdragonbae (approved):
lgtm

Issue #958, opened 2020-12-23
[QuickMenu] Implement QuickMenuStorageModel


jdragonbae:
We need a way to get button infos for quick menu. Currently, button infos hardcoded in controller.
It should be replaced with QuickMenuStorageModel.

PR #959, opened 2020-12-23
Implement QuickMenuStorageModel

jdragonbae:
This patch implements QuickMenuStorageModel. Furthermore, it will be
implemented to load and store data from persistent storage.

Refs: #958


src/org/triple/banana/quick_menu/QuickMenuStorageModelData.java
```diff
@@ -0,0 +1,15 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package org.triple.banana.quick_menu;
+
+import androidx.annotation.IdRes;
+import androidx.annotation.StringRes;
+
+class QuickMenuStorageModelData extends QuickMenuViewModelData.ButtonInfo {
```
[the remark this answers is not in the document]
jdragonbae: Thanks! Right, I agree with you. 
I'll update it.

src/org/triple/banana/quick_menu/QuickMenuFactoryImpl.java
```diff
@@ -14,9 +14,11 @@
     public @NonNull QuickMenuController create(@NonNull Context context) {
         MockQuickMenuViewImpl mockView = new MockQuickMenuViewImpl(context);
         QuickMenuViewImpl view = new QuickMenuViewImpl(context);
+        QuickMenuStorageModel storageModel = new QuickMenuStorageModelImpl();
```
codeimpl, line 17: Do we need storageModel variable?

jdragonbae: Hmm.. I think so. Did you mean inject instance directly? 
like `new QuickMenuControllerImpl(..., new QuickMenuStorageModelImpl() ...);`
Am I right?

codeimpl: yes, I think we don't need the variable anymore.

romandev: @codeimpl, Why do you think of that? Do you think that it's better to reduce the stack usage here?
IMHO, in this case, using variable might be better for readability than reducing stack usage. WDYT?

codeimpl: I thought creating unused variables was not a good way, but in terms of readability, the current state seems to be better.

romandev: @codeimpl Sorry, I don't understand what you mean exactly. Do you mean the one-time used variable? (but you mentioned unused variable)

codeimpl: Oh sorry. I mean the one-time used variable right. : ) BTW, I agree with you.

romandev (commented):
@jdragonbae Can you rebase this patch?
@codeimpl Is it okay to merge this as-is?

codeimpl:
> @jdragonbae Can you rebase this patch?
> @codeimpl Is it okay to merge this as-is?

sure~

src/org/triple/banana/quick_menu/QuickMenuFactoryImpl.java
jdragonbae: @romandev, @codeimpl Thanks for the review.

romandev (approved):
lgtm

Issue #978, opened 2020-12-29
Fix a build error in a case of 64bit APK


romandev:
```
DEX contains references to non-existent symbols after R8 optimization.
Tip: Build with:
        is_java_debug=false
        treat_warnings_as_errors=false
        enable_proguard_obfuscation=false
     and then use dexdump to see which class(s) reference them.

     E.g.:
       third_party/android_sdk/public/build-tools/*/dexdump -d out/Release/apks/YourApk.apk > dex.txt

Classe(s) without definition:
  Lorg/chromium/android_webview/common/PlatformServiceBridgeImpl;
  Lorg/chromium/weblayer_private/GmsBridgeImpl;
Field(s) without definition:
Method(s) without definition:
  Lorg/chromium/android_webview/common/PlatformServiceBridgeImpl;<init>()V
  Lorg/chromium/weblayer_private/GmsBridgeImpl;<init>()V
```

PR #979, opened 2020-12-29
Fix a build error in 64bit APK

romandev:
This is a regression after m88 rebase (#945). This is because the
following upstream change:
  - https://chromium-review.googlesource.com/c/chromium/src/+/2523532
So, this patch adds the error cases to exceptional list. Also, this
patch disables the weblayer. Because our triple_banana doesn't use it.

Refs: #978


jdragonbae (approved):
LGTM

PR #981, opened 2020-12-30
[WIP] Implement NewQuickMenuButton

romandev:


src/org/triple/banana/quick_menu/NewQuickMenuButton.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public NewQuickMenuButton(
+            Context context, AttributeSet attrs, int defStyleAttr, int defStyleRes) {
+        super(context, attrs, defStyleAttr, defStyleRes);
+    }
+
+    @Override
+    public void setCompoundDrawables(Drawable left, Drawable top, Drawable right, Drawable bottom) {
+        // FIXME: We should make a background for the "top" drawable.
```
romandev, line 33: @codeimpl I already mentioned it in offline discussion. We can make a background using `LayerDrawable` and then pass it to `super.setCompoundDrawables()`

Issue #975, opened 2020-12-28
Make the QuickMenuView item beautiful


jdragonbae:
To resolve this issue, we should make the QuickMenu UI so beautiful. If someone is not satisfied, we can't resolve this issue.

- [x] Make the image color of button dark mode applicable.
- [x] Add a ripple effect on the button.
- [x] Arrange the layout.

PR #982, opened 2020-12-30
Change default color for QuickMenuButton

codeimpl:
This patch applies default color for each icon because the buttons
should be changed automatically according to the mode (Dark or Light).

Refs: #975


romandev (approved):
lgtm

Issue #976, opened 2020-12-28
Refactor the QuickMenu module


jdragonbae:
To resolve this issue, we should make the roles of the QuickMenuViewModel and QuickMenuViewModel.Data clear.

PR #984, opened 2020-12-30
Implement generic Model class

romandev:
This is a generic model container class to define custom model data.
How to define the custom model data:
```
class UserDefinedModelData implements Model.Data {
    int hello = 1;
    ...
}
```

How to use that:
```
// Create a new model with |UserDefinedModelData|
Model<UserDefinedModelData> model = new
Model<>(UserDefinedModelData::new);

// Add a listener to listen the model data changed event
model.addListener(changedData -> {
    doSomething(changedData);
});

// Update the model data (but it is actually not updated yet before
// commit() is called)
model.data.hello = 2;

// Swap the pending data with the internal data and then notify the
// change event to the listeners.
model.commit();
```

Refs: #976


jdragonbae (commented):
lgtm

jdragonbae (approved):

codeimpl (approved):
I think it is really nice code! thanks~
LGTM

PR #985, opened 2020-12-30
Use new generic Model class in quick_menu module

romandev:
This patch removes the legacy model implementation in quick_menu and
then use a new generic Model class instead.

Refs: #976


src/org/triple/banana/quick_menu/QuickMenuViewModel.java
```diff
@@ -5,18 +5,47 @@
...
-    interface Listener<T> {
-        void onUpdate(@NonNull T data);
+import org.triple.banana.base.model.Model;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+
```
jdragonbae, line 19: nit: Empty line

romandev: Done.

jdragonbae (approved):
lgtm

Issue #968, opened 2020-12-28
Fix a bug that the play button doesn't work in media remote


jdragonbae:
**Reproducible steps:**
1. Build latest version in Release mode
2. Install the release mode APK
3. Play any movie on fullscreen mode with Media Remote
4. "Pause" button click

**Expected behavior**
The "Pause/Play" button should work well.

**Actual behavior**
The "Pause/Play" button doesn't work

romandev:
We decided to revert the related upstream patch.

PR #986, opened 2020-12-30
Fix a bug the play state doesn't work in media remote

jdragonbae:
This patch fixes a bug the play state doesn't work in the media remote.
The reason why it doesn't work is proguard. To resolve this issue, we
exclude this enum class in proguard.flags.

Refs: #968


romandev:
Closed by #1033

PR #993, opened 2021-01-05
Implement QuickMenuStorage and use it

romandev:
This patch introduces a new class `QuickMenuStorage` to load the quick
menu buttons data from storage. The class is similar to the existing
`QuickMenuStorageModel` but it is simplified and shares the `ButtonInfo`
with `QuickMenuViewModel`.

Refs: #976


jdragonbae (approved):
lgtm

PR #994, opened 2021-01-05
Use Runnable instead of QuickMenuActionProvider.Action

romandev:
There is no reason to define the custom `Action` class. It is duplicated
with `Runnable`.

Refs: #976


jdragonbae (approved):
LGTM

codeimpl (commented):
LGTM

PR #996, opened 2021-01-05
Update quick menu top layout

codeimpl:
Quick menu top layout include icon, app name, version. This patch
implemnts new top layout.

Refs: #975


romandev (approved):
lgtm

PR #997, opened 2021-01-05
Fix a bug that the quick menu doesn't work in debug build

romandev:
The problem is becuase the `QuickMenuStorageImpl` is not serializable.

Refs: #976


jdragonbae (approved):
lgtm

src/org/triple/banana/base/model/Model.java
```diff
@@ -95,6 +96,7 @@ private boolean wasDataChanged(@NonNull byte[] oldData, @NonNull byte[] newData)
             byte[] serializedData = baos.toByteArray();
             if (serializedData != null) return serializedData;
         } catch (Exception e) {
+            Log.e(TAG, "serialize(): " + e.toString());
```
jdragonbae, line 99: `import android.util.Log;` seems to be needed.

Issue #1006, opened 2021-01-06
[QuickMenu] Reduction quick menu vertical length


codeimpl:
How to
  - Reduction padding size of between icon and background 
  - Make two line as one line

To resolve 
  - [x] Reduction padding size
  - [x] Make two string line as one line
  - [x] Get consent for all members

PR #1016, opened 2021-01-09
[QuickMenu] Reduction vertical length of QuickMenu

codeimpl:
The QuickMenu dialog covers most of the entire screen. This patch
reduce the size of the dialog to improve usability.

How to
  - Reduction icon padding padding size
  - Make two string line as one line

Refs: #1006


romandev (approved):
lgtm

Issue #1007, opened 2021-01-06
[QuickMenu] When rotating, the screen should be well visible


codeimpl:
When changing the horizontal or vertical mode, the dialog screen should not be broken. To resolve this issue, there should be no problem with screen switching.

PR #1017, opened 2021-01-09
[QuickMenu] Support dialog rotation

codeimpl:
When dialog is rotating, child view on dailog was broken. So the dialog
adds RotationManager to get notify of system orientation change.

Refs: #1007


src/org/triple/banana/quick_menu/QuickMenuDialog.java
```diff
@@ -89,17 +98,47 @@ private void setColumnWeight(@NonNull QuickMenuMiddleButton button) {
         button.setLayoutParams(params);
     }
 
+    private void changeColumnCount(int columnCount) {
```
romandev, line 101: I think the better name might be `relayout`

src/org/triple/banana/quick_menu/QuickMenuDialog.java
```diff
@@ -89,17 +98,47 @@ private void setColumnWeight(@NonNull QuickMenuMiddleButton button) {
...
+    @Override
+    public void dismiss() {
+        super.dismiss();
+
+        mRotationManager.removeListener(mOrientationChangedListener);
+    }
+
+    private RotationManager.Listener createOrientationChangedListener() {
+        return info -> {
+            mHandler.post(() -> setGridLayoutColumn());
```
romandev, line 141: FYI, the `info` has already the orientation information. It means that the `setGridLayoutColumn()` might be unnecessary.

src/org/triple/banana/quick_menu/QuickMenuDialog.java
codeimpl: sure, thanks. done!

src/org/triple/banana/quick_menu/QuickMenuDialog.java
```diff
@@ -89,17 +98,43 @@ private void setColumnWeight(@NonNull QuickMenuMiddleButton button) {
...
+
+        mRotationManager.removeListener(mOrientationChangedListener);
+    }
+
+    private RotationManager.Listener createOrientationChangedListener() {
+        return info -> {
+            mHandler.post(() -> {
+                relayout(info.orientation == RotationManager.Orientation.LANDSCAPE
+                                ? Configuration.ORIENTATION_LANDSCAPE
+                                : Configuration.ORIENTATION_PORTRAIT);
```
romandev, line 136: There is no reason to convert the `RotationManager.Orientation.XXX` to `Configuration.ORIENTATION_XXX`

codeimpl: relayout function is called by onContentChanged() and here. But inside the onContentChanged only know orientatino of Configuration. So I thinks that it needs converting.

codeimpl: @romandev Could you check my code again?? thanks!

romandev (approved):
lgtm

Issue #1021, opened 2021-01-13
[QuickMenu] Make ViewModel Listener WeakReference


romandev:
In the current quick menu implementation, the View listens to the data changes from the ViewModel. So, the ViewModel has a strong reference to Listeners. However, the listeners can be implemented inside View. In this case, it causes the memory leaks because the ViewModel and View don't know each other so they can't remove the registered listener themselves.

PR #1022, opened 2021-01-13
Make the ViewModel.addListener take WeakReference

romandev:
In the current quick menu implementation, the View listens to the data
changes from the ViewModel. So, the ViewModel has a strong reference to
Listeners. However, the listeners can be implemented inside View. In
this case, it causes the memory leaks because the ViewModel and View
don't know each other so they can't remove the registered listener
themselves.
So, this patch makes the ViewModel.addListener take WeakReference and
also stores WeakReference instead of strong reference.

Refs: #1021


jdragonbae (approved):
lgtm

PR #1024, opened 2021-01-14
Fix a crash when a user clicks quick menu button

jdragonbae:
This patch fixes a crash when a user clicks quick menu button. To
resolve this issue, add null check for `@Nullable` parameter.

Refs: #1013


romandev (approved):
lgtm

src/org/triple/banana/quick_menu/QuickMenuMiddleButton.java
```diff
@@ -47,6 +47,9 @@ public void setCompoundDrawablesWithIntrinsicBounds(@Nullable Drawable left,
             @Nullable Drawable top, @Nullable Drawable right, @Nullable Drawable bottom) {
         Drawable background = ResourcesCompat.getDrawable(getResources(),
                 R.drawable.quick_menu_button_image_background, getContext().getTheme());
+        if (top == null || background == null) {
+            return;
```
romandev, line 51: I think we should call `super.setCompoundDrawablesWithIntrinsicBounds()` here

romandev (changes requested):

src/org/triple/banana/quick_menu/QuickMenuMiddleButton.java
jdragonbae: Ack. Thanks for the review.

romandev (approved):
lgtm

Issue #1012, opened 2021-01-06
[Toolbar] Migration existing user data for Banana(QuickMenu) button


codeimpl:
The banana button was fixed to the sixth through team consensus, and the existing user's toolbar is changed as follows to resolve this issue.
  1. If there is a banana button on the toolbar, it moves to the 6th position.
  2. If there is no banana button on the toolbar, replace it with the 6th button.

Issue #1011, opened 2021-01-06
[Toolbar] Make the Banana button uneditable


codeimpl:
As Banana button turns into a QuickMenu role, Banana button should always be fixed on the toolbar. To resolve this issue, Banana button should always be located on the bottom toolbar 6th.

PR #1025, opened 2021-01-15
[Toolbar] Fix QuickMenu button on BottomToolbar

codeimpl:
With the new implemenation of the QuickMenu button, it was decided to
pin it to the bottom toolbar. So, this patch have been modified the
following features.
  1. Migration the existing old user's QuickMenu button to the 6th
position
  2. Fixed the position of the 6th button

Refs: #1011, #1012


src/org/triple/banana/toolbar/SelectedToolbarButtonAdapter.java
```diff
@@ -187,7 +187,9 @@ public void setButtonList(ArrayList<ToolbarButtonItem> mButtonList) {
     public int getItemCount() {
         if (mButtonList == null) return 0;
 
-        return mButtonList.size();
+        // The total of the button list is six, but QuickMenu button should be fixed at 6th.
+        // So it should exclude the button lists.
```
romandev, line 191: The number of the toolbar buttons should be 6, but the 6th button should be the quick menu button. So, we need to exclude it from the total button count.

romandev (commented):
lgtm with nit

src/org/triple/banana/toolbar/SelectedToolbarButtonAdapter.java
codeimpl: done. thanks!

romandev (approved):
lgtm

PR #1033, opened 2021-01-18
Fix a bug that the play state is not updated in media remote

romandev:
It is a side effect after enabling vertical class merging in R8. So,
this patch simply reverts the related commit[1] in upstream side.

[1] c5c64c9365cac103bb58ab835598af4f3b6b1edc

Refs: #968


jdragonbae (commented):
lgtm

jdragonbae (approved):
lgtm

Issue #974, opened 2020-12-28
Decide the order of the items for QuickMenu


jdragonbae:
To resolve this issue, we should make an item list and the order of the items.

jdragonbae:
We discussed this issue via group call. the layout will be implemented through the reference.

romandev:
```
LAYOUT 
                                                            Clear data | Setting
—————————————————————————————-
AD BLOCK    | Secure DNS  |   Archive       | Visit history
Add to home | Download  |   DarkMode  | Find in Page
Print               |   QR code   |      Share       |  Desktop site
——————————————————————————————
 SECRET TAB   |           POWER             |      NEW TAB |
```

PR #1034, opened 2021-01-18
Fix a build error in ButtonInfoStorageImpl

romandev:
This is because of #1032.

Refs: #974, #1032


romandev:
TBR: @jdragonbae, @codeimpl

Issue #1038, opened 2021-01-19
Prevent showing the quick menu duplicately


romandev:
We should prevent showing the quick menu duplicately.

PR #1045, opened 2021-01-19
Prevent quick menu dialog to be shown multiple times

jdragonbae:
This patch prevents quick menu dialog to be shown multiple times.
Before showing the dialog, check whether the dialog is showing or not.

Refs: #1038


romandev (commented):
looks good but let's discuss about this PR again at offline

src/org/triple/banana/quick_menu/View.java
```diff
@@ -7,4 +7,5 @@
 
 interface View extends ViewModelBase.Listener {
     void show();
+    boolean isShowing();
```
codeimpl, line 10: Just question, do we need it?

romandev (approved):

Issue #1053, opened 2021-01-20
Separate dark mode preference from the user Interface


jdragonbae:
We should separate dark mode preference from the user interface. Then, it will be moved to the advanced features.

PR #1056, opened 2021-01-20
Separate the dark mode preference from the user interface

jdragonbae:
This patch separates the dark mode preference from the user interface.
And, this is moved to the advanced features.

Refs: #1053


romandev (approved):
lgtm

Issue #1054, opened 2021-01-20
Replace clear data with user interface on quick menu


jdragonbae:
Currently, the clear data button is on the quick menu top layout. As we discussed on the offline meeting, we decided to replace clear data with user interface.

PR #1058, opened 2021-01-20
Replace clear data with user interface on quick menu

jdragonbae:
This patch replaces clear data with user interface on quick menu.

Refs: #1054


romandev (approved):
lgtm

Issue #1049, opened 2021-01-20
Implement translate feature


jdragonbae:
To resolve this issue, implement translate feature.
When a user triggers translate, the popup for translate should be shown.

PR #1060, opened 2021-01-20
Implement translate button on quick menu

jdragonbae:
This patch implements translate button on quick menu. Since there is a
limit on the number of quick menu buttons, we decided to replace add to
home with this.

Refs: #1049


upstream/tb_components_translate_core_browser_translate_manager.cc.patch
```diff
@@ -0,0 +1,31 @@
+diff --git tb/components/translate/core/browser/translate_manager.cc tb_diff/components/translate/core/browser/translate_manager.cc
+--- tb/components/translate/core/browser/translate_manager.cc
++++ tb_diff/components/translate/core/browser/translate_manager.cc
+@@ -265,6 +265,7 @@ bool TranslateManager::CanManuallyTranslate(bool menuLogging) {
+   // Translation of unknown source language pages is supported on desktop
+   // platforms, but not mobile.
+ #if defined(OS_ANDROID) || defined(OS_IOS)
++#if !defined(ENABLE_TRIPLE_BANANA)
```
romandev, line 8: So, if we enable this code by force, the mobile web page is actually translated? or not?

romandev (commented):
lgtm

upstream/tb_components_translate_core_browser_translate_manager.cc.patch
jdragonbae: The mobile web page is actually translated in spite of `unknown language`.

romandev (approved):
lgtm

Issue #921, opened 2020-12-13
[Secure DNS] Enable the Secure DNS mode without restarting browser


jdragonbae:
Currently, we need to restart the browser to use Secure DNS mode and Bypass. 
We should make this enabled without restarting.

PR #1062, opened 2021-01-22
Enable CECPQ2 feature without restarting browser

jdragonbae:
This patch enables CECPQ2 feature without restarting the browser. When
secure DNS mode is turn on, enable CECPQ2 feature in a network.

Refs: #921


romandev (commented):
I'm not sure but I think this patch can't cover the whole cases such as the packet breaking. We need an ability to set the flag at runtime.

jdragonbae:
> I'm not sure but I think this patch can't cover the whole cases such as the packet breaking. We need an ability to set the flag at runtime.

Ack. Thanks for the review. As we discussed, I'll try to show the restart dialog when the secure dns mode is changed.

PR #1064, opened 2021-01-25
Show restart dialog when the secure dns mode is changed

jdragonbae:
This patch shows the restart dialog when the secure dns mode is changed.
When a user turn on/off the secure dns switch preference, we inform the
user that the browser needs to be restarted for applying the secure dns
feature.

Refs: #921


jdragonbae:
We decide to enable this flag without restart. A patch for this issue will be uploaded.

Issue #1065, opened 2021-01-26
Fix a crash when the secure dns preference is clicked


jdragonbae:
To resolve this issue, we should investigate the below crash log.

````
01-26 16:15:46.225 11292 11292 E AndroidRuntime: FATAL EXCEPTION: main
01-26 16:15:46.225 11292 11292 E AndroidRuntime: Process: org.triple.banana, PID: 11292
01-26 16:15:46.225 11292 11292 E AndroidRuntime: java.lang.RuntimeException: Unable to start activity 
ComponentInfo{org.triple.banana/org.chromium.chrome.browser.settings.SettingsActivity}: 
Bi0: Unable to instantiate fragment org.banana.cake.interfaces.BananaSecureDnsSettings: make sure class name exists
````

PR #1070, opened 2021-01-26
Fix a crash when a user clicks the secure dns preference

jdragonbae:
This patch fixes a crash when a user clicks the secure dns preference.
According to the change[1] in upstream, the fragment must have an empty
constructor to keep the class.

[1] https://chromium-review.googlesource.com/c/chromium/src/+/2036872

Refs: #1065


romandev (approved):
lgtm

Issue #1036, opened 2021-01-19
Adjust the animation time to show the quick menu


romandev:
We should make it faster than now.

PR #1076, opened 2021-01-31
Adjust the animation time to show the QuickMenu

codeimpl:
The animation times are sometimes fast, sometimes slow. Therefore, the
time is not cosistent. Root cause is the animation was canceled as the
layout was changed. This patch changed to complete layout inflating
before showing the dialog.

Refs: #1036


romandev (approved):
LGTM

Issue #1078, opened 2021-02-01
Fix a release build error


jdragonbae:
This is regression about #1066.

To resolve this issue, investigate the error log.

```
DEX contains references to non-existent symbols after R8 optimization.
Tip: Build with:
        is_java_debug=false
        treat_warnings_as_errors=false
        enable_proguard_obfuscation=false
     and then use dexdump to see which class(s) reference them.

     E.g.:
       third_party/android_sdk/public/build-tools/*/dexdump -d out/Release/apks/YourApk.apk > dex.txt

Classe(s) without definition:
  Landroidx/preference/PreferenceFragmentCompat;
Field(s) without definition:
Method(s) without definition:
  Landroidx/preference/PreferenceFragmentCompat;requireActivity()Landroid/app/Activity;

Traceback (most recent call last):
  File "../../build/android/gyp/proguard.py", line 676, in <module>
    main()
  File "../../build/android/gyp/proguard.py", line 663, in main
    options.warnings_as_errors)
  File "../../build/android/gyp/proguard.py", line 464, in _CheckForMissingSymbols
    fail_on_output=warnings_as_errors)
  File "/home/supreme/Desktop/android_chromium/src/build/android/gyp/util/build_utils.py", line 292, in CheckOutput
    raise CalledProcessError(cwd, args, MSG.format(stream_string))
util.build_utils.CalledProcessError: Command failed: ( cd /home/supreme/Desktop/android_chromium/src/out/release64; ../../third_party/jdk/current/bin/java -Xmx1G -cp ../../third_party/r8/lib/r8.jar com.android.tools.r8.tracereferences.TraceReferences --map-diagnostics:MissingDefinitionsDiagnostic error warning --check --lib ../../third_party/android_sdk/public/platforms/android-30/optional/org.apache.http.legacy.jar --lib ../../third_party/android_system_sdk/android_system.jar --source obj/triple_banana/triple_banana/triple_banana.r8dex.jar )
Command failed because it wrote to stderr.
You can often set treat_warnings_as_errors=false to not treat output as failure (useful when developing locally).
ninja: build stopped: subcommand failed.
```

PR #1075, opened 2021-01-31
Disable bytecode_rewriter_target for androidx_fragment

jdragonbae:
Disable bytecode_rewriter_target for androidx_fragment

This patch disables bytecode_rewriter_target flag for androidx_fragment
aar. According to the upstream[1], this flag modify the return type of
AndroidX's Fragment.getActivity() method from FragmentActivity to
Activity. It causes the problem[2] in android biometricx aar which used
for our secure login feature. As a stopgap, disable this flag.

[1] https://source.chromium.org/chromium/chromium/src/+/master:docs/ui/android/bytecode_rewriting.md

[2] https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:biometric/biometric/src/main/java/androidx/biometric/BiometricFragment.java;l=262

Refs: #1066

PR #1079, opened 2021-02-02
Fix a release build error

jdragonbae:
This patch fixes a release build error about #1075 regression. As the
bytecode_rewriter_target flag is turned off, the code that uses it
should be modified. So, we should manually build the jar file[1] and
then replace the existing one. Additionally, this jar file seems to have
been removed from the current upstream.

[1] third_party/android_deps/local_modifications/androidx_preference_preference/androidx_preference_preference_java.jar

Refs: #1075, #1078


romandev (approved):
lgtm

Issue #1077, opened 2021-02-01
Add basic settings preference in banana extension


jdragonbae:
To resolve this issue, add the basic settings preference in banana extension. When it's clicked, launch the upstream setting activity.

PR #1080, opened 2021-02-02
Add browser settings preference in banana extension

jdragonbae:
This patch adds the browser settings preference in banana extension.
It is belong to the other settings category. When it's clicked, launch
the upstream setting activity.

Refs: #1077


banana_cake/interfaces/java/org/banana/cake/interfaces/BananaToolbarManager.java
```diff
@@ -34,4 +38,5 @@ static BananaToolbarManager get() {
     void openQRCodeDialog();
     boolean getUseDesktopUserAgent();
     void translateCurrentTab();
+    void openBrowserSettings(@NonNull Context context);
```
romandev, line 41: It looks this is not actually used in toolbar. So, we should move this to `BananaApplicationUtils`

jdragonbae: Thanks for the review. I was confused. I'll move it.

romandev (approved):
lgtm

Issue #1081, opened 2021-02-03
Fix a bug that could not showing quick menu when tab is empty


jdragonbae:
To resolve this issue, we should implement `getLastFocusedActivity` api then use it.

PR #1084, opened 2021-02-03
Add getLastTrackedFocusedActivity api to applitcation utils

jdragonbae:
This patch adds getLastTrackedFocusedActivity api to the
BananaApplicationUtils. We can use this to get the last focused
activity. Also, this patch fixes a bug that could not showing quick menu
when tab is empty by using activity instead of tab activity.

Refs: #1081


src/org/triple/banana/quick_menu/QuickMenuServiceImpl.java
```diff
@@ -31,7 +31,7 @@ public void show(@NonNull Context context) {
         ButtonStateManager stateManager = ButtonStateManager.get();
         ViewController controller =
                 new ViewControllerImpl(viewModel, storage, actionProvider, stateManager);
-        View view = new DialogBasedViewImpl(context, controller);
+        View view = new DialogBasedViewImpl(activity, controller);
```
romandev, line 34: Can we change the signature of DialogBasedViewImpl's constructor as well?

jdragonbae: Thanks for the review. Done.

romandev (approved):
lgtm

Issue #1085, opened 2021-02-03
The update_upstream command doesn't support binary changes


romandev:
Currently, our `update_upstream` command to generate upstream patch files couldn't support any binary change. So far this hasn't been a problem, but after the change[1] for build error, it actually becomes a problem. Because always generates wrong diffs during update_upstream.

[1] https://github.com/triplebanana/triple_banana/pull/1079

PR #1088, opened 2021-02-03
Add an exception to be execluded during update_upstream command

romandev:
Currently, our update_upstream command to generate upstream patch files
couldn't support any binary change. So far this hasn't been a problem,
but after the change[1] for build error, it actually becomes a problem.
Because always generates wrong diffs during update_upstream.
So, this patch adds an exception to be excluded during update_upstream
command.

[1] https://github.com/triplebanana/triple_banana/issues/1085

Refs: #1079, #1085


jdragonbae (approved):
lgtm

PR #1091, opened 2021-02-03
Reorder terminate and add secret button

codeimpl:
We decided to reorder terminate and add secret button because of add
secret button will be used more often than the terminate button.

Refs: none


jdragonbae:
LGTM with nit.

nit : I think we used `Refs: none` instead of `Refs: None`.

codeimpl:
> LGTM with nit.
> 
> nit : I think we used `Refs: none` instead of `Refs: None`.

done! :)

jdragonbae (approved):
lgtm

romandev (commented):
Hmm, as far as I know, we decided to keep the current order. Am I missing something?

codeimpl:
As I remember.. We decide to reorder only adding secret tab and power. Am I wrong??;;;;

romandev:
We decided to drop this patch again.

Issue #1083, opened 2021-02-03
Fix a top icon area on the quick menu layout


jdragonbae:
Currently, a top icon ( User Interface, Banana extension) is small so that a user mistake easily when click it. We can expand the padding size of top icon to fix it.

PR #1092, opened 2021-02-03
[QuickMenu] Increase touch area for top icon

codeimpl:
Touch area of top icon is small to click. This patch increase the touch 
area to avoid touch miss.

Refs: #1083


romandev (approved):
lgtm

PR #1093, opened 2021-02-04
Handle CECPQ2 feature flag at runtime without restarting process

romandev:
Currently, there is no way to enable the CECPQ2 feature even if Secure
DNS is enabled. Moreover, it should have restarted the browser to enable
the CECPQ2 feature. So, this patch adds a new method `SetEnabled()` to
`FeatureList` and handle the CECPQ2 feature flag at runtime without
restarting process.

Refs: #921


jdragonbae (approved):
lgtm

PR #1095, opened 2021-02-05
Add "Reset adblock filter" preference

romandev:
This patch adds "Reset adblock filter" preference explictly in
AdblockFeatureSettings.

Refs: #927


jdragonbae (approved):
LGTM

Issue #1096, opened 2021-02-06
Secure DNS data migration doesn't work correctly


romandev:
**Reproducible steps**
1. Install 5.x apk
2. Enable "Secure DNS" feature
3. Upgrade 6.x apk
4. Disable "Secure DNS" feature
5. Restart browser

**Actual behavior**
Secure DNS is always enabled in 6.x apk after upgrade even if users disable the feature.

**Expected behavior**
Secure DNS should be disabled if users disable the feature.

PR #1097, opened 2021-02-06
Fix a bug that the Secure DNS data migration doesn't work correctly

romandev:
Currently, there is a bug that the Secure DNS is always enabled after
upgrade from 5.x to 6.x even if users disable the feature. It is because
ExtensionFeatures.isEnabled(SECURE_DNS) is always true after upgrade.
So, this patch make the flag off after data migration.

Refs: #1096


jdragonbae (approved):
lgtm

PR #1098, opened 2021-02-07
Fix a crash in debug build

romandev:
`Abort message: '[FATAL:feature_list.cc(371)] Check failed: !GetInstance()->initialized_. '`

Refs: none


jdragonbae (approved):
lgtm

PR #1099, opened 2021-02-07
Ignore sentinel file mechanism

romandev:
The ruleset indexer is using `sentinel file` concept, which is used as a
safeguard to prevent crash-looping if ruleset indexing crashes right
after start-up.
The sentinel file is placed in the ruleset version directory just before
indexing commences, and removed afterwards. Therefore, if a sentinel
file is found on next start-up, it is an indication that the previous
indexing operation may have crashed, and indexing will not be attempted
again.
Admittedly, this approach errs on the side of caution, and false
positives can happen. For example, the sentinel file may fail to be
removed in case of an unclean shutdown, or an unrelated crash as well.
As a result, it causes subresource filter broken sometimes. So, this
patch simply ignores the sentinel file until the problem is getting
fixed.

Refs: #927


romandev:
TBR: @codeimpl, @jdragonbae

PR #1102, opened 2021-02-08
Fix a bug that the QR code scan doesn't work if there are no tab

romandev:
There is a bug that the QR code scan feature doesn't work if there are
no tab. It is because openQRCodeDialog() returns early if the active tab
is absent. So, this patch fixes the problem by making the QRCodeDialog
open even if there are no active tab.

Refs: #886


jdragonbae (approved):
lgtm

PR #1103, opened 2021-02-08
Fix a crash when clicking search button after dark mode change

romandev:
Currently, CakeToolbarManager internally caches an instance of the
toolbar manager in main activity. The instance of the toolbar manager
might already be gone when the activity is recreated after dark mode
change.

Refs: none


jdragonbae (approved):
lgtm

PR #1106, opened 2021-02-08
Fix a blinking bug in the banana extension settings

jdragonbae:
This patch fixes a blinking bug in the banana extension settings. When
onResume is called, we don't need a inflating the layout.

Refs: none


romandev (approved):
lgtm

PR #1107, opened 2021-02-08
Fixed toolbar editor layout issue

codeimpl:
The bottom toolbar editor layout was broken do to the increase of the
buttons. This patch just reduce height of vertical margin.

Refs: none


romandev (approved):
lgtm

PR #1109, opened 2021-02-09
Fix a bug that the Secure DNS migration is sometimes failed

romandev:
Secure DNS migration is sometimes failed when launching app from
application launcher while updating APK. This patch fixes the problem by
introducing wasDnsModeSet() API.

Refs: #1096


romandev:
TBR: @jdragonbae, @codeimpl

PR #1124, opened 2021-03-16
Rebase 89.0.4389.86

romandev:
This patch makes our browser use Chromium m89 but we still have two
unresolved issues as follows:
  - The initCommandLine() should be called in somewhere (#1131)
  - The bottom toolbar doesn't appear since m89 rebase (#1132)

Refs: #1123


jdragonbae (approved):
lgtm

PR #1126, opened 2021-03-23
Impl/set loop

jdragonbae:


src/org/triple/banana/media/MediaCommandProcessor.java
```diff
@@ -84,4 +84,8 @@ public void setRelativeVolume(float value) {
     public void setVideoFit(@NonNull MediaFitState state) {
         runMediaCommand("style.objectFit = '" + state + "'");
     }
+
+    public void setLoop(@NonNull MediaLoopState state) {
+        runMediaCommand("loop = " + state);
```
romandev, line 89: I'm not sure but isn't the `loop` is `boolean`?

src/org/triple/banana/media/MediaLoopState.java
```diff
@@ -0,0 +1,26 @@
...
+
+public enum MediaLoopState {
+    NOT_SUPPORTED,
+    LOOP,
+    NOT_LOOP;
+
+    @Override
+    public @NonNull String toString() {
+        final MediaLoopState state = values()[ordinal()];
+        return String.valueOf(state == LOOP);
```
romandev, line 18: Ah, okay. I understand now why you write the following code:
`runMediaCommand("loop = " + state);`

But I don't think it is good because the code isn't intutive.
We might be able to make the `isLoop()` as follows
```
public boolean isLoop() {
    return ordinal() == LOOP.ordinal();
}
```
But do we really need it? We could just check as follows:
`runMediaCommand("loop = " + (state == MediaLoopState.LOOP));`

src/org/triple/banana/media/MediaLoopState.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    public @NonNull MediaLoopState toggle() {
+        final MediaLoopState state = values()[ordinal()];
+        if (state == NOT_SUPPORTED) return NOT_SUPPORTED;
+        return state == LOOP ? NOT_LOOP : LOOP;
```
romandev, line 24: For readability.
```
switch (state) {
    case LOOP: return NOT_LOOP;
    case NOT_LOOP: return LOOP;
    default:
        assert state == LOOP || state == NOT_LOOP;
        return NOT_SUPPORTED;
}
```

src/org/triple/banana/media/MediaLoopState.java
jdragonbae: Looks more simple to me. Thanks for the reivew. I'll change it.

src/org/triple/banana/media/MediaLoopState.java
jdragonbae: Done

Issue #1120, opened 2021-03-16
Implement getting playing options


jdragonbae:
To resolve this issue, we should implement getting playing options(loop or not).

PR #1127, opened 2021-03-24
Implement getting loop state for media remote

jdragonbae:
This patch implements getting loop state for media remote. When the loop
state is set, we could dispatch this event. According to this state, set
the view model to change the view.

Refs: #1120


upstream/tb_third_party_blink_renderer_core_dom_element.cc.patch
```diff
@@ -0,0 +1,29 @@
...
+ }
+ 
+ void Element::SetBooleanAttribute(const QualifiedName& name, bool value) {
++#if defined(ENABLE_TRIPLE_BANANA)
++  if (RuntimeEnabledFeatures::MediaRemoteEnabled() &&
++      name == html_names::kLoopAttr) {
++    FullscreenMediaController::Get().DispatchLoopStateChanged(
++        value ? media::mojom::BananaLoopState::LOOP
++              : media::mojom::BananaLoopState::NOT_LOOP);
++  }
```
romandev, line 25: This affects all elements. It's not good in performance perspective. IMHO, we can update the loop state when updating current time and duration.

Issue #1116, opened 2021-03-16
Add a icon for video fit


jdragonbae:
To resolve this issue, add a icon for video fit. It should be added on the media remote.

PR #1128, opened 2021-03-25
Add fit button icons on the media remote

jdragonbae:
This patch adds fit button icons for the media remote. A user can select
video fit mode.

Refs: #1116


res/drawable/ic_fit_cover.xml
```diff
@@ -0,0 +1,16 @@
...
+     file, You can obtain one at https://mozilla.org/MPL/2.0/. -->
+
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path android:pathData="M0 0h24v24H0V0z" />
+    <path
+        android:fillColor="@android:color/white"
```
romandev, line 14: I'm  not sure but shouldn't we use default color for dark mode?

res/drawable/ic_fit_fill.xml
```diff
@@ -0,0 +1,16 @@
...
+     file, You can obtain one at https://mozilla.org/MPL/2.0/. -->
+
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path android:pathData="M0 0h24v24H0V0z" />
+    <path
+        android:fillColor="@android:color/white"
```
romandev, line 14: ditto

res/drawable/ic_fit_cover.xml
jdragonbae: Thanks for the review. In my opinion, I think it would be a white color when in dark mode. Cause, if it has a dark color, it has a little bit hard to visible on the media remote. WDYT?

romandev (approved):
lgtm

Issue #1118, opened 2021-03-16
Implement setting video fit state


jdragonbae:
To resolve this issue, should implement executing javascript which set 'object-fit' property on video element.

PR #1129, opened 2021-03-25
Implement setting video fit state

jdragonbae:
This patch implements setting video fit state for media remote.
And, it adds MediaFitState to define state on the view model.

Refs: #1118


src/org/triple/banana/media/MediaFitState.java
```diff
@@ -0,0 +1,12 @@
+// Copyright 2021 The Triple Banana Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package org.triple.banana.media;
+
+public enum MediaFitState {
```
romandev, line 8: If you can write some utility method in Enum class, then you can get rid of a lot of switch statements.
```
    enum MediaFitState {
        FILL,
        COVER,
        CONTAIN;

        @Override
        public @NonNull String toString() {
            return super.toString().toLowerCase();
        }

        public static @NonNull MediaFitState from(@NonNull String value) {
            try {
                return MediaFitState.valueOf(value.toUpperCase());
            } catch (Exception e) {
                return CONTAIN;
            }
        }

        public @NonNull MediaFitState toggle() {
            return values()[(ordinal() + 1) % values().length];
        }
    }
```

src/org/triple/banana/media_remote/MediaRemoteService.java
```diff
@@ -189,6 +190,21 @@ public void onMediaRemoteButtonClicked(int id) {
...
+                    break;
+                case COVER:
+                    fitState = MediaFitState.CONTAIN;
+                    break;
+                case CONTAIN:
+                default:
+                    fitState = MediaFitState.FILL;
+                    break;
+            }
+            mMediaController.setVideoFit(fitState);
```
romandev, line 207: mMediaController.setVideoFit(mViewModel.getData().getFitState().toggle());

src/org/triple/banana/media/MediaCommandProcessor.java
```diff
@@ -79,4 +80,21 @@ public void setRelativeVolume(float value) {
...
+                break;
+            case COVER:
+                fitState = "cover";
+                break;
+            case CONTAIN:
+            default:
+                fitState = "contain";
+                break;
+        }
+        runMediaCommand("style.objectFit = '" + fitState + "'");
```
romandev, line 98: runMediaCommand("style.objectFit = '" + state + "'");

src/org/triple/banana/media/MediaFitState.java
jdragonbae: Thanks for the review! Sounds good to me.

src/org/triple/banana/media_remote/MediaRemoteService.java
jdragonbae: Done.

src/org/triple/banana/media/MediaCommandProcessor.java
jdragonbae: Done.

romandev (approved):
lgtm

Issue #1117, opened 2021-03-16
Implement getting video fit states


jdragonbae:
To resolve this issue, we should implement getting video fit states (fill, cover, contain).

PR #1130, opened 2021-03-26
Implement getting video fit state

jdragonbae:
This patch implements getting video fit state for media remote.

Refs: #1117


upstream/tb_diff_third_party_blink_renderer_core_fullscreen_fullscreen_media_controller.cc.patch
```diff
@@ -76,8 +77,15 @@
 +
 +  SetVideoElement(video_element);
 +  EnsureBindEventDispatcher();
++
++  CSSStyleDeclaration* style = video_element->style();
++  String object_fit("");
++  if (style)
++    object_fit = style->GetPropertyValueInternal(CSSPropertyID::kObjectFit);
++
 +  dispatcher_->OnEnteredVideoFullscreen(processor_.BindNewPipeAndPassRemote(),
```
romandev, line 86: We don't want to make a change in native side.

jdragonbae: Thanks for the review. I removed the change in the native side.

modules/public/mojom/media.mojom
```diff
@@ -16,7 +16,7 @@ interface BananaMediaEventDispatcher {
   OnPlayStateChanged(BananaPlayState state);
   OnTimeUpdate(double current_time, double duration);
   OnChangedPIPMode(bool value);
-  OnEnteredVideoFullscreen(pending_remote<BananaMediaCommandProcessor> impl, bool isDownloadable);
+  OnEnteredVideoFullscreen(pending_remote<BananaMediaCommandProcessor> impl, bool isDownloadable, string fitState);
```
romandev, line 19: I think you could use enum (like BananaPlayState) with enum traits. But I'm okay with as-is if it is difficult and complex.

src/org/triple/banana/media/MediaEventDispatcher.java
```diff
@@ -63,10 +65,24 @@ public void onChangedPipMode(boolean value) {
...
+                state = MediaFitState.FILL;
+                break;
+            case "cover":
+                state = MediaFitState.COVER;
+                break;
+            case "contain":
+            default:
+                state = MediaFitState.CONTAIN;
+                break;
+        }
```
romandev, line 82:
```
MediaFitState.from(state);
```

src/org/triple/banana/media/MediaCommandProcessor.java
```diff
@@ -97,4 +103,8 @@ public void setVideoFit(@NonNull MediaFitState state) {
         }
         runMediaCommand("style.objectFit = '" + fitState + "'");
     }
+
+    public void getVideoFit(@NonNull BananaJavaScriptCallback callback) {
```
romandev, line 107: This is unnecessary

src/org/triple/banana/media/MediaCommandProcessor.java
```diff
@@ -16,14 +17,19 @@
 
     private @Nullable BananaMediaCommandProcessor mImpl;
 
-    private void runMediaCommand(String mediaCommand) {
+    private void runMediaCommand(@NonNull String mediaCommand) {
+        runMediaCommandWithCallback(mediaCommand, null);
+    }
+
+    private void runMediaCommandWithCallback(
```
romandev, line 24: This is unnecessary becasue we decided to drop using the JS way.

src/org/triple/banana/media/MediaController.java
```diff
@@ -63,4 +64,8 @@ public void setRelativeVolume(float value) {
     public void setVideoFit(@NonNull MediaFitState state) {
         mCommandProcessor.setVideoFit(state);
     }
+
+    public void getVideoFit(@NonNull BananaJavaScriptCallback callback) {
```
romandev, line 68: Unnecessary

src/org/triple/banana/media_remote/MediaRemoteService.java
```diff
@@ -205,6 +208,23 @@ public void onMediaRemoteButtonClicked(int id) {
                     break;
             }
             mMediaController.setVideoFit(fitState);
+            mMediaController.getVideoFit(result -> {
```
romandev, line 211: Unnecessary

src/org/triple/banana/media_remote/MediaRemoteService.java
```diff
@@ -205,6 +208,23 @@ public void onMediaRemoteButtonClicked(int id) {
...
+                final MediaFitState state;
+                switch (result.replace("\"", "")) {
```
romandev, line 213: Just query: Why do we need to replace this string here?

upstream/tb_diff_third_party_blink_renderer_core_fullscreen_fullscreen_media_controller.cc.patch
```diff
@@ -76,8 +77,15 @@
 +
 +  SetVideoElement(video_element);
 +  EnsureBindEventDispatcher();
++
++  CSSStyleDeclaration* style = video_element->style();
++  String object_fit("");
++  if (style)
++    object_fit = style->GetPropertyValueInternal(CSSPropertyID::kObjectFit);
```
romandev, line 84: Is this your intention? IMHO, if the `object-fit` is set by user style sheet (.css), this value might be always empty even if the computed value is NOT empty. You need to check computed value using `GetComputedValue()`.

src/org/triple/banana/media/MediaEventDispatcher.java
jdragonbae: Done

src/org/triple/banana/media/MediaCommandProcessor.java
jdragonbae: Thanks! Done

src/org/triple/banana/media/MediaCommandProcessor.java
jdragonbae: Thanks! Done

src/org/triple/banana/media_remote/MediaRemoteService.java
jdragonbae: Done

src/org/triple/banana/media/MediaController.java
jdragonbae: Done

modules/public/mojom/media.mojom
jdragonbae: Thanks for the guidance.

src/org/triple/banana/media_remote/MediaRemoteService.java
jdragonbae: Because the JavaScriptCallback result would be with a double quotation like `"contain"` not `contain`.

upstream/tb_diff_third_party_blink_renderer_core_fullscreen_fullscreen_media_controller.cc.patch
jdragonbae: Oh, I see. Thanks for the review! I'll change it.

upstream/tb_diff_third_party_blink_renderer_core_fullscreen_fullscreen_media_controller.cc.patch
```diff
@@ -76,8 +77,10 @@
 +
 +  SetVideoElement(video_element);
 +  EnsureBindEventDispatcher();
-+  dispatcher_->OnEnteredVideoFullscreen(processor_.BindNewPipeAndPassRemote(),
-+                                        video_element->SupportsSave());
++
++  dispatcher_->OnEnteredVideoFullscreen(
++      processor_.BindNewPipeAndPassRemote(), video_element->SupportsSave(),
++      WebElement(video_element).GetComputedValue("object-fit"));
```
romandev, line 83: The GetComputedValue() might return `WebString()`. Then, does it works well? You might have to check whether it actually works well. I'm not sure but it wouldn't work.

jdragonbae: When I test it, it seems to be worked. But if it would be a worry point I willing to change it. WDYT?

romandev: > When I test it, it seems to be worked. But if it would be a worry point I willing to change it. WDYT?

Does it mean that you tried to make the `GetComputedValue()` return WebString() always?

Issue #1131, opened 2021-03-29
The initCommandLine() should be called in somewhere


romandev:
The `initCommandLine()` is missed after m89 because the `onAttachContext()` is moved to `SplitCompatApplication`. We should call the `initCommandLine()` in somewhere.

romandev:
Failure upstream files:
```
upstream/tb_chrome_android_java_src_org_chromium_chrome_browser_ChromeApplication.java.patch
```

PR #1133, opened 2021-03-30
Introduce a new way of Cake API

romandev:
This patch introduces/implements a new way of Cake API. The new API
replaces part of the existing Banana API (called from upstream code).
The new way fixes the dependency problem of the existing API.

```
--------------------      -----------------      --------------
|Banana App Module | <--- | Cake API Impl | ---> | Banana API |
--------------------      -----------------      --------------
|Chrome Java       |              |
--------------------              |
|Individual module |              |
--------------------              |
|Cake API Interface| <|------------
--------------------
```

Refs: #1131, #1136


romandev:
@jdragonbae @codeimpl gentle ping

android/src/banana/cake.java
```diff
@@ -0,0 +1,8 @@
+package banana;
```
jdragonbae, line 1: Can we add a copy right?

jdragonbae (approved):
lgtm

Issue #1139, opened 2021-04-02
A crash occur in debug build


romandev:
```
04-03 02:48:24.619 14771 14771 D AndroidRuntime: Shutting down VM
04-03 02:48:24.620 14771 14771 E AndroidRuntime: FATAL EXCEPTION: main
04-03 02:48:24.620 14771 14771 E AndroidRuntime: Process: org.triple.banana_beta, PID: 14771
04-03 02:48:24.620 14771 14771 E AndroidRuntime: java.lang.ExceptionInInitializerError
04-03 02:48:24.620 14771 14771 E AndroidRuntime:        at org.chromium.chrome.browser.base.SplitCompatApplication.attachBaseContext(SplitCompatApplication.java:160)
04-03 02:48:24.620 14771 14771 E AndroidRuntime:        at org.banana.cake.bootstrap.BananaApplication.attachBaseContext(BananaApplication.java:22)
04-03 02:48:24.620 14771 14771 E AndroidRuntime:        at android.app.Application.attach(Application.java:370)
04-03 02:48:24.620 14771 14771 E AndroidRuntime:        at android.app.Instrumentation.newApplication(Instrumentation.java:1157)
04-03 02:48:24.620 14771 14771 E AndroidRuntime:        at android.app.LoadedApk.makeApplication(LoadedApk.java:1288)
04-03 02:48:24.620 14771 14771 E AndroidRuntime:        at android.app.ActivityThread.handleMakeApplication(ActivityThread.java:7152)
04-03 02:48:24.620 14771 14771 E AndroidRuntime:        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:7134)
04-03 02:48:24.620 14771 14771 E AndroidRuntime:        at android.app.ActivityThread.access$1600(ActivityThread.java:274)
04-03 02:48:24.620 14771 14771 E AndroidRuntime:        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2102)
04-03 02:48:24.620 14771 14771 E AndroidRuntime:        at android.os.Handler.dispatchMessage(Handler.java:107)
04-03 02:48:24.620 14771 14771 E AndroidRuntime:        at android.os.Looper.loop(Looper.java:237)
04-03 02:48:24.620 14771 14771 E AndroidRuntime:        at android.app.ActivityThread.main(ActivityThread.java:8167)
[...]

PR #1140, opened 2021-04-03
Fix a crash when launching our browser

romandev:
It is caused by IllegalAccessException during instantiateReflectively()
in Module class because the CakeApiImpl class is declared as package
private. So, this patch makes the CakeApiImpl public.

Refs: #1139


jdragonbae (approved):
lgtm

